PR #12, opened 2018-09-06
Enable unused variables warning

rachitnigam:
@tedbauer I'm enabling stricter checks for the project. The build system throws a few unused variable warnings now. Can you build this branch and change any intentional unused variables to `_`?

### DIFF
--- a/src/emit.ml
+++ b/src/emit.ml
@@ -141,17 +141,12 @@ let rec emit_cmd i cmd =
   | CFuncDef (id, args, body)      -> emit_fun (id, args, body) i
   | CApp (id, args)                -> emit_app (id, args) i
   | CTypeDef (id, t)               -> emit_typedef (id, t) i
-  | CMuxDef _        -> emit_muxdef
+  | CMuxDef _                      -> ""
 
 and emit_assign_int (id, e) =
   concat [ "int "; id; " = "; (emit_expr e); ";" ]
 
-<<<<<<< HEAD
 and emit_assign_arr (id, _, d) i =
-=======
-(* FIXME(rachit): [e] is unused. Remove it from param list or use it. *)
-and emit_assign_arr (id, e, d) i =
->>>>>>> c1b794ad77d439e39b941fb48e3fd637ca37f61d
   let bf = compute_bf d in
   let arr_size = compute_array_size d in
   let part_pragma =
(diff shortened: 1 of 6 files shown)


tedbauer:
Ok, I believe I've just changed all the unused variables to "_". There are also a lot of non-exhaustive pattern matching, would you like me to fix that?

src/emit.ml
```diff
@@ -141,12 +141,17 @@ let rec emit_cmd i cmd =
...
 and emit_assign_int (id, e) =
   concat [ "int "; id; " = "; (emit_expr e); ";" ]
 
+<<<<<<< HEAD
 and emit_assign_arr (id, _, d) i =
+=======
+(* FIXME(rachit): [e] is unused. Remove it from param list or use it. *)
+and emit_assign_arr (id, e, d) i =
+>>>>>>> c1b794ad77d439e39b941fb48e3fd637ca37f61d
   let bf = compute_bf d in
```
rachitnigam: ^ Fix please.

src/emit.ml
```diff
@@ -215,7 +220,11 @@ and emit_typedef (id, t) i =
   concat [ "typedef "; (type_str t); " "; id; ";" ]
   |> indent i
 
+<<<<<<< HEAD
 and emit_muxdef _ = "" (* No need to emit anything *)
+=======
+and emit_muxdef = "" (* No need to emit anything *)
+>>>>>>> c1b794ad77d439e39b941fb48e3fd637ca37f61d
```
rachitnigam: Fix please

src/type.ml
```diff
@@ -145,20 +142,21 @@ and check_for_impl id r1 r2 body u (ctx, delta) =
     else raise (TypeError range_static_error)
   | _ -> raise (TypeError range_error)
 
+(* TODO(rachit): [d] is unused. Is this a mistake? *)
 and check_assignment id exp (ctx, delta) =
-  check_expr exp (ctx, delta) |> fun (t, (c, d)) ->
```
rachitnigam: Are you sure this is correct? If so, can you add a comment explaining why?

src/type.ml
```diff
@@ -127,10 +124,10 @@ and check_for id r1 r2 body (ctx, d) =
   | _ -> raise (TypeError range_error)
 
 and check_for_impl id r1 r2 body u (ctx, delta) =
-  check_expr r1 (ctx, delta) |> fun (r1_type, (c1, d1)) ->
```
rachitnigam: Are you sure this is correct? The definition for `check_for` has different logic.

rachitnigam:
Yeah, might as well. You can just add cases saying `failwith "Not yet implemented"`.

PR #21, opened 2018-09-10
Flatten sequence representation

### LINKED ISSUE #17: CSeq should take a list of commands
A common optimization for the `seq` representation is to have it take a list of commands instead of a tuple of two. This keeps the AST flat and reduces the chance of stack overflows.

tedbauer:
Fixes #17.

### DIFF
--- a/src/parser.mly
+++ b/src/parser.mly
@@ -1,5 +1,13 @@
 %{
 open Ast
+
+let parse_sequence c1 c2 =
+  match c1, c2 with
+  | CSeq cmds1, CSeq cmds2 -> CSeq (cmds1 @ cmds2)
+  | c, CSeq cmds           -> CSeq (c :: cmds)
+  | CSeq cmds, c           -> CSeq (cmds @ [c])
+  | c1, c2                 -> CSeq ([c1; c2])
+
 %}
 
 %token EOF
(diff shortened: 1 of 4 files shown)


sampsyo:
That's a good point about the `c1 = cmd; c2 = cmd` rule! I don't know either how that's guaranteed to match. It's worth experimenting, but if you're curious about an alternative, I think a slightly more standard way to do this would be to introduce a new nonterminal just for sequences. That would look like `statement_list` in [this SO answer](https://stackoverflow.com/a/26868625), whose only purpose in life is to "left-recurse" and eventually ground out in `statement`.

[This section in RWO](https://v1.realworldocaml.org/v1/en/html/parsing-with-ocamllex-and-menhir.html#parsing-sequences) also has some relevant advice.

tedbauer:
I'll try that out! Thanks Adrian!

src/parser.mly
```diff
@@ -61,7 +61,15 @@ cmd:
     LBRACK c = cmd RBRACK
     { CFor (x, x1, x2, 1, c) } ;
   | c1 = cmd c2 = cmd
-    { CSeq (c1, c2) }
```
rachitnigam: To respond to the question about which case happens: menhir currently resolves this in an arbitrary manner. If you want to make it more obvious, we should change the rule to be `c1 = cmd SEMICOLON c2 = cmd` and remove `SEMICOLON` from other rules. Additionally, we define `%right SEMICOLON` so that only the second case happens.

rachitnigam: This is done in a better way in https://github.com/cucapra/seashell/commit/6ad10c48ec36235c31b280dc8714827296fa4df5

rachitnigam:
LGTM

PR #28, opened 2018-09-19
Implement capabilities for arrays using linear types

rachitnigam:
WIP

### DIFF
--- a/bin/seac.ml
+++ b/bin/seac.ml
@@ -3,35 +3,14 @@ open Cmdliner
 open Seashell
 open Compile_utils
 
-let seac filename no_typecheck print_ast : unit =
+let seac filename print_ast : unit =
   Printexc.record_backtrace false;
   let prog = In_channel.read_all filename in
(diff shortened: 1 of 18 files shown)


tedbauer:
[Here](https://github.com/cucapra/seashell/blob/cap-2/src/infer_cap.ml)'s an attempt at implementing read-capability inference (I still need to add back in write, but I think it should be way simpler than read to add back in). Concerned it might be inefficient/too ad hoc and gross, but I couldn't think of a better way to do it. The gist of the implementation is this:

- For any expressions in `CApp`s, `CIf`s, `CReassign`s, `CAssign`s, and `CFor`s, check for any subexpressions that require read capabilities (namely `EAA`s and `EBankedAA`s), and then transform these commands to `CSeq`s containing any necessary read capabilities and then finally the old commands.
- Maintain a context that keeps track of expressions that already have capabilities generated for them, because we don't want to generate multiple.

PR #34, opened 2018-10-10
Arbitrary precision integers, wip

tedbauer:
Contains support for the following:

- When writing an int literal in a Seashell program, you can write `number<nb>` to specify that it should be represented with `b` bits. If you just write `number`, 32 bits is assumed. Example: `10<5b>` is the integer 10, represented with 5 bits.
- Generated C code contains `#include<apcint.h>`, and int types will include the number of bits, as per chapter 2 in [this](https://www.xilinx.com/support/documentation/sw_manuals/xilinx2018_2/ug902-vivado-high-level-synthesis.pdf).
- Typechecking does this: only allow operations between integers with overlapping ranges. The type of the result has the range of the larger-ranged integer.


test/type_failures.ml
```diff
@@ -5,7 +5,7 @@ open Test_utils
 let%expect_test "cannot add float and int" =
   compile_string_with_failure "let x = 2.5; let z = 3; x + z;";
   [%expect {|
-    [Type error] can't apply operator + to float and idx<3..4, 0..1>. |}]
+    [Type error] can't apply operator + to float and idx<3..4, -2147483648..2147483647>. |}]
```
rachitnigam: Why did this change the dynamic range of z?

test/type_failures.ml
```diff
@@ -20,12 +20,12 @@ let%expect_test "Cannot access non-existent bank" =
...
   compile_string_with_failure "func foo(a: int[10], x: bool) { a[1] := x; }";
   [%expect {|
-    [Type Error] cannot assign value of type `bool' to L-value of type `idx<0..1, -4611686018427387904..4611686018427387903> lin'. |}]
+    [Type Error] cannot assign value of type `bool' to L-value of type `idx<0..1, -2147483648..2147483647> lin'. |}]
 
 let%expect_test "Cannot reassign to different type" =
   compile_string_with_failure "let x = 2.5; x := 1;";
   [%expect {|
-    [Type Error] cannot assign value of type `idx<1..2, 0..1>' to L-value of type `float'. |}]
+    [Type Error] cannot assign value of type `idx<1..2, -2147483648..2147483647>' to L-value of type `float'. |}]
```
rachitnigam: Same question as above.

test/should-compile/ap_ints.sea
```diff
@@ -0,0 +1,12 @@
+/////////////////////////////////////////////
+// ap_ints.sea
```
rachitnigam: You need to add a test to the test_programs.ml file in order to get this test to run.

rachitnigam:
For the first point, I thought we decided that instead of having explicitly sized numbers, we’ll assume that any static number is represented in (log2 n) bits.

(I assume you’re working on this, but as a reminder to myself) emit needs to emit the right types and seashell needs a way to specify these types.

PR #35, opened 2018-10-16
Arbitrary precision integers v2!

tedbauer:
Does this:
- Instead of parsing `int` in type annotations, parses `bit<n>`.
- When there are operations between single-value index types (i.e. index types with trivial static components), the resulting index type has a dynamic component that envelops both of the operands' dynamic ranges.
- The type strings in the error messages for index types now say `bit<n>` when appropriate.

Also updated the test files to have `bit<32>` annotations instead of `int`.

Closed the other PR and just did this in a fresh new branch.

**edit**: now emitted code includes `apcint.h` header and adds number of bits to `int` type annotations.

### DIFF
--- a/test/type_failures.ml
+++ b/test/type_failures.ml
@@ -46,3 +46,9 @@ let%expect_test "Cannot shadow variables" =
   compile_string_with_failure "func foo(a: bit<32>[10]) { let a = 10; }";
   [%expect {|
     `a' already bound in this context. Cannot shadow variables. |}]
+
+let%expect_test "Cannot reassign int to something stored with more bits" =
+  compile_string_with_failure "func foo(a: bit<32>, b: bit<33>) { a := b; }";
+  [%expect {|
+    [Type Error] Cannot reassign value of type `unsigned bit<32>' to value of type `unsigned bit<33>';
+         `unsigned bit<32>' is represented with less bits than `unsigned bit<33>'.|}]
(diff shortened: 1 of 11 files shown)


test/type_failures.ml
```diff
@@ -8,41 +8,41 @@ let%expect_test "cannot add float and int" =
...
+  compile_string_with_failure "func foo(a: bit<32>[10], x: bit<32>) { a{1}[1]; }";
   [%expect {|
     [Type error] Bank 1 already consumed for memory `a' |}]
 
 let%expect_test "Cannot assign incorrect type to array" =
-  compile_string_with_failure "func foo(a: int[10], x: bool) { a[1] := x; }";
+  compile_string_with_failure "func foo(a: bit<32>[10], x: bool) { a[1] := x; }";
   [%expect {|
-    [Type Error] cannot assign value of type `bool' to L-value of type `idx<0..1, -4611686018427387904..4611686018427387903> lin'. |}]
+    [Type Error] cannot assign value of type `bool' to L-value of type `unsigned bit<32> lin'. |}]
```
rachitnigam: Nice!

rachitnigam:
Add some tests to show that type checker fails in the right cases.

rachitnigam:
Also seems like you're missing the emit code.

tedbauer:
Oh, oops! I'll add that in right now.

PR #40, opened 2018-11-15
Views, wip

tedbauer:
Includes:
- View parsing, typechecking and compiling

Todo:

- [ ] More testing!
  - [X] Create simple view, access view in a loop
  - [X] Create view with dynamic offset, access view in nested loop
  - [X] Create multidimensional view, access in doubly nested loop
  - [x] Create view on a view
  - [x] Create MD view on a view
  - [ ] Create more complicated views
  - [X] Type error example with oversized view (width of view > size of underlying array)
  - [x] Type error example with view made from with an iterator that's unrolled
  - [x] Type error example with view 
- [X] Multidimensional semantics
- [X] Some cleanup

Parses view expressions, then the type checker pass treats the views like arrays, so the checking is simple. Then, a separate "view resolving" pass after the type checking converts accesses on views to accesses on the original arrays in the AST, and then the code emitting is unchanged.


src/ast_visitor.ml
```diff
@@ -61,6 +62,8 @@ class ['s] ast_mapper = object(self)
     let e1', st1 = self#expr e1 st in
     let e2', st2 = self#expr e2 st1 in
     EBankedAA (id, e1', e2'), st2
+  method private eview (id, params) st =
+    EView (id, params), st
```
rachitnigam: This seems incorrect. You would want to visit the `expr`s in the list with `elist_visit` method and use that to recreate the `EView`.

try
```diff
@@ -0,0 +1,6 @@
+type int = bit<10>;
```
rachitnigam: This file looks like a spurious additon.

test/type_failures.ml
```diff
@@ -56,3 +56,9 @@ let%expect_test "Cannot reassign int to something stored with more bits" =
 let%expect_test "Non-existent type" =
   compile_string_with_failure "func foo(a: non_type[10]) { let x = 2; }";
   [%expect {| [Type Error] No type definition for `non_type'.; |}]
+
```
rachitnigam: Can you add the following tests:
- The view is created from a non-array.
- The view is created using an iterator that is being unrolled.

rachitnigam:
Can you resolve the conflicts with master?

Overall this looks good. I'll hold off on merging this till we figure out the correct mux based constraints. I don't think they should cause too much rewriting.

tedbauer:
Sounds good, will do.

try
tedbauer: Whoops! Deleted.

Issue #37, opened 2018-10-23
Sized ints overflow


tedbauer:
For the current implementation of sized integers, huge numbers are stored in the index types, and with bit numbers as high as 32 or so there are overflow errors. It seems a bit silly to store such big numbers, so maybe we could change the `TIndex` representation for the implementation? Would need to think about how to differentiate static ints I think.

sampsyo:
Good point. One trick might be to store, for the range endpoints, _either_ a plain number _or_ a power of two:

```
type Offset = Lin of int | Exp of int
```

so `Exp 32` represents the number 2^32 while `Lin 32` represents plain old 32.

rachitnigam:
@tedbauer is this fixed?

rachitnigam:
@tedbauer ping

rachitnigam:
Implemented in #46

PR #43, opened 2019-01-29
Index overflow fix

tedbauer:
This addresses issue #37. Instead of having the range endpoint of the dynamic component be a potentially enormous number, it's something of type `offset`, which is either a normal number, or an exponential number (`Lin of int | Exp of int`).


rachitnigam:
Fixed in #46

PR #48, opened 2019-02-06
scala: Use a simpler command-line syntax

sampsyo:
This doesn't require `-f` to pass the file to parse; it's now a positional argument, so you type `./fuse file.sea` to compile. I also used scopt's `File` option type, which does some path-specific stuff.


rachitnigam:
Lgtm

PR #52, opened 2019-02-07
Capabilities implementation

### LINKED ISSUE #50: Capabilities in unrolled contexts
An array access needs to be able to generate sufficient resources to be consumed in an unrolled context:
```
decl a: bit<10>[3 bank 3][3 bank 3];
for (i ...) unroll 3 {
  for (j ...) unroll 3 {
     a[i][j] // correct
     ---
     a[0][j]
     ---
     a[j][i] //correct
  }
}
```
A non-unrolled context implicitly unrolls by `1` which means constants are allowed since they generate exactly one copy of the access:
```
for(i ...) {
  a[0] // correct
  for (j ...) unroll 3 {
    a[0][j] //correct I think?
    a[j][0] // correct I think?
    a[i][j] // correct I think
    a[i][0] // wrong
  }
}
```

rachitnigam:


rachitnigam:
**TLDR**: Reasoning about read/write capabilities should be performed at the source level, not the manually unrolled semantics.

I think this is what @sa2257 is asking about:
```
a: int[10]
for (i ... ) {
  for (j ... ) unroll 3 {
    let v = a[i][0] // should work since it turns into a multi-read
  }
}
```
@sa2257 is saying that this should be fine since it turns into:
```
a: int[10]
for (i ... ) {
  pfor (j ...) {
    let v1 = a[i][0]
    let v2 = a[i][0]
    let v3 = a[i][0]
  }
}
```
However, this line of reasoning is wrong. Consider a simpler example:
```
for (i ... ) unroll 3 {
  let v = a[0] // should work by the same line of reasoning
} combine {
  x += v
}
```
For this example to work, the `k`-register `v` would have to read `a[0]` **three times same cycle** which shouldn't be possible (@sa2257 is it?).

The right way to formulate read/write capabilities is through this example:
```
a: int[9 bank 3]
for (i ...) unroll 3 {
  let v = a[i];
  let x = a[i]
}
```
the multiple read are correct in `for` because they are **syntactic multi-reads**, i.e. they can be turned into an assignment to temp register that writes into `v` and `x`. 

@sampsyo comments?

sampsyo:
I think the answer depends on whether you imagine the language has implicit or explicit capabilities. Imagine two different versions of the language: one (massively inconvenient) version (A) where you have to explicitly acquire and then use capabilities, and another (B) where the language looks more like an ordinary imperative language. In language A, you need to use `let r = a[i][j].read(); let value = r();` to actually read anything, whereas in language B, you can just say `let value = a[i][j];`.

In language A, I would say that _both_ of the first two code blocks above (with and without `unroll`) should be ruled out. In language A, you would have to write this program:

```
a: int[10]
for (i ... ) {
  let readcap = a[i][0].read();
  pfor (j ...) {
    let v1 = readcap();
    let v2 = readcap();
    let v3 = readcap();
  }
}
```

This is valid, as would an analogous adaptation of the first (`unroll`-containing) example above.

Of course, we want people to write language B instead. And that makes for a more complicated type rule, but whatever we use should be sound with respect to a language-B desugaring. Does that sound right?

I didn't quite follow your last code listing, which is quite different because there's no two-level loop nest—maybe you can expand on what you meant to ask about that example?

rachitnigam:
For the language A rewrite, does the fact that you lifted `a[i][0]` to the outer loop matter. Should I be able to write this in language A:
```
for (i ...) {
  pfor (j ...) {
    let readcap = a[i][0].readcap()
    let v1 = readcap();
    let v2 = readcap()
    let v3 = readcap()
  }
}
```

If the answer is no, should language B allow for listing 1 to be correct?

rachitnigam:
FWIW, I think lifting `readcap` or `v` in language A or B make the underlying hardware design much more explicit. I would prefer if you had to manually lift it out.

sampsyo:
1. I say no: in language A, acquiring a capability _always_ consumes the appropriate banks, regardless of whether it's a read or write capability. So that is just as illegal (for the same reason) as this one:

    ```
    for (i ...) {
      pfor (j ...) {
        let writecap = a[i][0].write();
        writecap(5);
      }
    }
    ```

    …whose body corresponds to `a[i][0] := 5` in language B.

2. First, it matters less whether this is allowed in language B, because either choice is sound. (Unlike the first question, in which only one answer is sound.) Allowing more convenient indexing (without promoting reads) is an issue of convenience, not correctness. I tend to agree with you that, for now at least, a more explicit version of the language will help clarify the type system and make the implementation simpler. If we drive ourselves crazy doing the manual LICM to satisfy the type system, maybe we can revisit the convenience issue.

rachitnigam:
Okay. The current implementation requires LICM. 

However, I'm not convinced that either choice is sound. Let me try out some rules on paper and see if it makes sense.

sampsyo:
It depends on the rule, certainly—I was saying that that _specific_ example where the program reads the same location 3 times could be allowed. But the hard part would be to design a rule that allows that program while still ruling out all the bad programs.

rachitnigam:
Yes, exactly. The naive rule that comes to mind is very weak and might allow wrong programs.

rachitnigam:
@sampsyo I've implemented a different algorithm to allow for read capabilities to be acquired inside unrolled contexts to allow for what @sa2257 was talking about.

The psuedocode for the algorithm is described on top of the TypeCheck file.

sampsyo:
Hmm… can we perhaps talk about this in person? I don't quite see the case that this is attempting to allow.

More broadly, I'm somewhat concerned about handing _capability acquisition_ differently for read and write capabilities. The original idea behind the capability system was to avoid exactly this distinction: the type system could handle acquisition uniformly, so getting a capability would _always_ consume the appropriate resources, regardless of whether you're writing or reading. (That step would work "linearly," i.e., capability acquisition requires the necessary banks as arguments and doesn't return them.) Then, once you have a read or write capability, those could be used differently—i.e., read capabilities can be used multiple times ("nonlinearly") whereas write capabilities cannot.

This design sounds like read capability _acquisition_ is being made nonlinear, so we don't have the uniformity on acquisition anymore. Is that right? If so, and if that's necessary, that's OK—but we probably should scrap the capability intermediary altogether and just check reads and writes directly. (And embrace language B instead of language A.)

rachitnigam:
I don't follow the comment about non-linear read capabilities.

The current implementation is much more like language B and everything is implemented in the type checker. When the type checker infers a read capability like this:
```
a[0][i] + x
```
It first linearly consumes the banks corresponding to `[0][i]` from `a` and provides a `Read -> a[0][i]` in the context. The next time `a[0][i]` is seen *in the same context*, the type checker first checks if a capability exists and if so, skips the bank consumption step.

The check is slightly different for writes. When a write is implied, the type checker consumes the relevant banks **and also** does the resource check corresponding to the unroll context (if inside a `unroll 3` followed by `unroll 5`, the access must allow for 15 copies). It then adds a `Write -> a[0][i]` to the context. However, if the same expression occurs in the context, it fails since multi-writes are not allowed. 

If in a new context, the read/write capability are removed but the consumed banks are still kept around which means this still errors correctly:
```
if (...) { a[0][i] := 1 };
a[0][i] := 2
```

rachitnigam:
Also more concretely, it is trying to allow for the matrix-vector multiply example I committed. If you have the original restriction where both read and writes check for the resource constraint, you cannot unroll both the loops (even with LICM).

sampsyo:
OK, cool—that helps a lot. At the risk of stating the obvious (but just for my own clarity), another way to phrase what's going on here is that the type checker is doing a CSE pass on array accesses first to insert the capability acquisition. So this:

```
f(a[0][i] + x);
g(a[0][i] + x);
```

becomes, after CSE:

```
let r = a[0][i].read();
f(r() + x);
g(r() + x);
```

And this is fine, because it's OK to reuse the read capability `r`. (This is what I mean by saying that read capabilities themselves can be used nonlinearly.)

On the other hand, this:

```
a[0][i] := x;
a[0][i] := y;
```

becomes, after CSE:

```
let w = a[0][I].write();
w(x);
w(y);
```

which is rejected, because write capabilities are linear.

As you said, the difference between read & write capability acquisition has to do with unrolling. Stealing an example from your Slack message, this is the proposal:

```
for (...) unroll 3 { a[0] + 1 }  // OK (unlike before)
for (...) unroll 3 { a[0] := 1 }  // error (as before)
```

Of course, manual LICM "works" for this example in the sense that it's possible to move the `a[0]` access out of the loop to avoid the need for the newly loosened policy. You mentioned that the impetus for the new relaxation is the matrix/vector multiplication example:

```
for (let i = 0..6) unroll 3 {
  for (let j = 0..8) unroll 4 {
    let rs = M[i][j] * V[j];
  } combine {
    R[i] += rs;
  }
 }
```

And I see now that the problem is the `V[j]` access, which clearly can't be LICM'd (it's not loop-invariant w.r.t. the inner loop!) but also violates the "unrolling restriction" in the sense that _writing_ to `V[j]` right there would certainly be a problem. So we need some other way to certify that this is OK.

The thing that worries me about getting rid of the "unrolling restriction" on reads altogether, however, is that it makes it harder to convince myself that we are ruling out all the bad problems. With that restriction in place, it's possible to imagine doing all the checking on the appropriately CSE'd form of the program, where I think it's pretty easy to argue that capability acquisitions ensure mutual exclusion on all banks. But with this change, I don't yet see the simplified form where this kind of argument can be made as airtight.

More forthcoming after I can ponder this for a little while…

rachitnigam:
So, @sa2257 claims that writing `V[j]` inside the inner loop is _not_ an issue since the hardware should be able to create multiple wires from each bank and send the read value into the multipliers. Is there another reason why you think `V[j]` is unsafe?

sampsyo:
No, I agree that it is safe (and that we should somehow allow that matrix/vector multiply program). I'm just worried about the simplicity of the type system. I don't have a counterexample showing that the new policy is unsound, but I'm also still not 100% convinced yet that it is sound (working on convincing myself of this now).

sampsyo:
Oh wait, sorry! You said "writing" `V[j]`.

Here's an example program:

```
for (let i = 0..6) unroll 3 {
  for (let j = 0..8) unroll 4 {
    V[j] := f(i, j);
  }
 }
```

This seems to write potentially-different values to `V[0]` simultaneously, for example, which seems like it should be rejected.

rachitnigam:
No, I meant "writing `V[j]` in the program". I agree the above example is wrong, but the read `V[j]` seems fine.

rachitnigam:
Also, just to be clear, the type system would reject the write to `V[j]` in that context.

sampsyo:
OK, here are the results of my pondering, which were a little easier to write up by hand: [on relaxing the unrolling restriction.pdf](https://github.com/cucapra/seashell/files/2849294/on.relaxing.the.unrolling.restriction.pdf)

I think the core insight is that `V[j]` in the above example is only safe because we get to assume the "copies" of the inner loop run in "lockstep." So the two copies of the loop simultaneously want to read `V[0]`, and then they simultaneously want `V[1]`, and so on.

The reason this gives me pause is that this means the safety is _only_ guaranteed on the version of the program with `unroll` in it, and we don't have a way to certify the safety for the "desugared" version of the program that copies n' pastes the loop bodies. Here's the simplified program from my PDF:

```
for (i...) unroll 2 {
  for (j...) {
    let x = a[j];
  }
}
```

(Note that the inner loop need not be unrolled to expose the effect here.) This is safe because the two copies of the inner loop run in lockstep. If we desugar the program in the usual way, however:

```
for (i...) {  // executing half as many iterations now
  for (j...) {
    let x = a[j];
  }
  for (j...) {
    let x = a[j];
  }
}
```

our current type rules would reject this program (even though it's equivalent to the original). And it's hard to see how we would design a straightforward system that would allow this latter program. The PDF contains an example of a slightly-different program that really should be rejected, because it makes the loops run "out of sync," i.e., not in lockstep.

My (second) somewhat nefarious example of a program that subtly ruins the lockstep iteration thing is:

```
for (i...) unroll 2 {
  f(i);
  // Do something where the (logical-time) latency depends on i:
  // for example, run a loop for i iterations.
  for (j...) {
    let x = a[j];
  }
}
```

Now the two "copies" of the `j` loop are no longer in lockstep, so they may want different indices of `a` at the same time.

That's the problem I'm worried about… and at the moment, I don't have a good solution for this. I'll need to continue pondering to have that!

rachitnigam:
You're right in saying that the desugared version is explicitly disallowed (see conversation with @sa2257 yesterday in slack) while the unroll example is allowed.

For the final example, if `f(i)` doesn't create dataflow dependencies, wouldn't it be scheduled in parallel with the inner for loop (or alternatively, it's safe to reorder it an move it below the inner for loop)?

sampsyo:
Good point—does it help to imagine a `---` after the call to `f(i)`?

rachitnigam:
That depends on the answer to this question: Are these programs equal?
```
for (...) unroll 3 {
C1
----
C2
}
```
```
for (... ) {
  C1; C1
  ---
  C2; C2
}
```
```
for (...) {
 { C1 --- C2 };
 { C1 --- C2 };
}
````

sampsyo:
Good question! We definitely haven't been thinking of any of those programs as equivalent, but perhaps in light of this discussion, we need to. Something along the lines of the first equivalence (between the first and second programs above), for example, seems necessary to preserve "lockstep" semantics.

rachitnigam:
Merging without a fix to #58.

Issue #62, opened 2019-02-13
Functions of some sort


tedbauer:
Some of the MachSuite apps are lengthy, like `backprop/backprop`, and it would be nice to organize Fuse programs into logical chunks. Would it be possible to have some sort of construct for this, if not functions?

sampsyo:
Indeed! See #54 for a proposal.

PR #67, opened 2019-02-13
Function support

### LINKED ISSUE #54: Functions implementation sketch
Proposal to bring back functions:

- Create new `CFuncDef` and `EApp` nodes to represent function definitions and applications. Change `Prog` to be a list of declarations, followed by a list of function definitions, followed by top level commands.
  - Alternatively, we can try removing `decl`s altogether and require a `main` function to be defined.
- For typechecking, extend the environment to map function identifiers to `CFuncDef` nodes. Checking `EApp`s is non-trivial since memories used in an `EApp` expression need have their banks consumed correctly. Follow this algorithm:
  - Conservatively assume that all banks in parameters are consumed.
  - ~Check of if the actual parameters satisfy constraints on formal parameters.~
  - ~Typecheck the function body using formal parameters and "write back" the bank consumptions to actual paramters.~
  - ~Optimization: Pre calculate these consumptions instead of re-running for each EApp.~
- Extend `Emit` to spit out function definitions.
- Recursive functions are **not allowed**.
- Use before define for functions is **not allowed**.

rachitnigam:
Fixes #54 #62.

- Program structure is now a list of function defs followed by a list of decls, followed by top level commands.
- Applications consume array expressions completely
- Functions cannot return values (will potentially fix this in the future) and have return type `void`.

### DIFF
--- a/src/main/scala/Emit.scala
+++ b/src/main/scala/Emit.scala
@@ -51,13 +51,16 @@ private class Emit extends PrettyPrinter {
   implicit def IdToString(id: Id): Doc = value(id.v)
 
   implicit def typeToDoc(typ: Type): Doc = typ match {
+    case _:TVoid => "void"
     case _:TBool | _:TIndex | _:TStaticInt => "int"
     case _:TFloat => "float"
     case _:TSizedInt => value(typ)
     case TArray(typ, _) => typ
+    case _:TFun => throw Impossible("Cannot emit function types")
   }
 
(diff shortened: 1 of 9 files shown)


rachitnigam:
I'll merge this EOD. Let me know if you have any questions.

sampsyo:
Yay! Looks perfect to me. I concur with all-`void` return values for now; seems like most of the functions we'll want to call will just mutate an input buffer.

Issue #57, opened 2019-02-11
Views w/ shrink and flex


rachitnigam:
The current formulation of views requires that the underlying array have the same number of banks as the width of the dimension of the view.

Examples:
(1) Should fail:
```
decl a: int[10 bank 10];
view v_a = a[off = 0, w = 5]; 
```

(2) Should pass
```
decl a: int[10 bank 10];
view v_a =  a{shrink 5}[off = 0, w = 5]
```
(3) Should pass
```
decl a: int[10 bank 10];
view v_a = a{flex 3}[off = 0, w= 3]
```

**Q1.** Should the programmer be allowed to freely mix `flex` and `shrink` in multidimensional cases? Which examples should pass:
(4)
```
decl a: int[10 bank 10][10 bank 10]
view v_a = a{shrink 5}{}[off = 0, w = 5][off = 0, w = 10]
```
(5)
```
decl a: int[10 bank 10][10 bank 10]
view v_a = a{shrink 5}{shrink 5}[off = 0, w = 5][off = 0, w = 5]
```
(6)
```
decl a: int[10 bank 10][10 bank 10]
view v_a = a{shrink 5}{flex 3}[off = 0, w = 5][off = 0, w = 3]
```

**Q2.** What the stepping semantics of `shrink` and `view`? Did we decide that the step of a `shrink` array is always equal to the length and that the steps for `flex` can be arbitrary?

**Q3.** Syntax suggestions for `shrink`, `flex`, and `views`?

sampsyo:
1. I don't see any immediate problems with any of these three—they all seem like they should probably be fine? It could help to try drawing the "hardware interposer" for each to make sure they seem to have roughly predictable cost.
2. I think those restrictions make sense, yes.
3. Some rough ideas, based on previous discussions:

        let v = view a[0..+5];  // v is a view on a with offset 0 and width 5
        let a' = shrink a: int[10 bank 5];   // a' is like a, but with banking factor 5
        let a' = shrink a: int[10 bank 5][10 bank 5];   // for multi-dimensional arrays

    `flex` could work the same way. The idea behind the `shrink` syntax is that it might be nice, for clarity, to write out the new type of the "adapted" memory.

rachitnigam:
Alternative proposal: If `shrink` forces the step size to be the same as the width, the compilation looks like: 
```
decl a: bit<10>[16 bank 8];
decl b: bit<10>[8];

for (let i = 0..8) {
  view v_a = shrink a[i:2]
[...]

PR #71, opened 2019-02-14
Implementation of shrink views

rachitnigam:
This PR implements `shrink` views and partially addresses #57.

Overview:
- Shrink view can be defined using `view v = a[0:w]` or `view v = a[n * i:w]`.
  - In the second case `n * off`, `n` is the step size and **must be equal** to `w`. This is to make sure that programmer does trip with the stepping semantics of `shrink` views.
- Shrink views cannot be defined in an unroll context. This restriction can potentially be removed in the future.
- For an array `a: t[len bank b]...`, the view must have the same number of dimensions, the same type `t` and banking factor `b'` such that `b % b' == 0`. This also means that `shrink` views are always fully banked.
- A `view v = <view_type> arr...` expression completely consumes `arr` from the current context.

**TODO**
- [X] Write a code gen phase.

### DIFF
--- a/src/main/scala/RewriteView.scala
+++ b/src/main/scala/RewriteView.scala
@@ -1,0 +1,109 @@
+package fuselang
+
+/**
+ * AST pass to rewrite views into simple array accesses. Should be used after
+ * type checking.
+ *
+ * For `shrink` views, we rewrite them as:
+ *
+ * view v_a = shrink a[4 * i : 4];
+ * v_a[k];
+ * ==>
+ * ; // remove the view decl
+ * a[4*i + k]
+ *
+ * If `a` itself is a view, we keep rewriting it until we reach a true array.
+ */
+object RewriteView {
+  import Syntax._
+  import CodeGenHelpers._
+
+	// We can use a flat environment because the type checker makes sure things
+  // in the right scope.
+  type T = Map[Id, List[Expr] => Expr]
+
+  def foldExprs(es: List[Expr])(implicit env: T): (List[Expr], T) =
+      es.foldLeft((List[Expr](), env))({
+        case ((idxsn, env), e) => {
+          val (e1, env1) = rewriteExpr(e)(env)
+          (e1 :: idxsn, env1)
+        }
+      })
+
+  def rewriteExpr(e: Expr)(implicit env: T): (Expr, T) = e match {
+    case EVar(_) | EInt(_) | EFloat(_) | EBool(_) => e -> env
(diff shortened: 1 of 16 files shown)


rachitnigam:
@sampsyo ready for review. Questions/comments?

src/main/scala/RewriteView.scala
```diff
@@ -0,0 +1,109 @@
...
+ * ; // remove the view decl
+ * a[4*i + k]
+ *
+ * If `a` itself is a view, we keep rewriting it until we reach a true array.
+ */
+object RewriteView {
+  import Syntax._
+  import CodeGenHelpers._
+
+	// We can use a flat environment because the type checker makes sure things
```
sampsyo: Errant indentation?

src/main/scala/RewriteView.scala
```diff
@@ -0,0 +1,109 @@
...
+  // in the right scope.
+  type T = Map[Id, List[Expr] => Expr]
+
+  def foldExprs(es: List[Expr])(implicit env: T): (List[Expr], T) =
```
sampsyo: Having "fold" in the name here is a little weird because this isn't really a fold operation (because it's not higher order). Of course, perhaps having a general `fold` for expressions could be useful?

src/main/scala/RewriteView.scala
```diff
@@ -0,0 +1,109 @@
...
+      val (e1n, env1) = rewriteExpr(e1)
+      val (e2n, env2) = rewriteExpr(e2)(env1)
+      cr.copy(lhs = e1n, rhs = e2n) -> env2
+    }
+    case CExpr(exp) => {
+      val (e1n, env1) = rewriteExpr(exp)
+      CExpr(e1n) -> env1
+    }
+    case CEmpty => c -> env
+  }
```
sampsyo: Seeing this does sort of convince me that an abstracted `fold` for AST transformers could be useful…

rachitnigam: What’s the type you’re thinking of?

sampsyo: I suppose I actually meant a `map` for ASTs! So the type would be something like `map(expr: Expr, f: Expr => Expr): Expr`. It could take care of making the recursive calls and re-creating tree nodes after applying `map(subtree, f)` to their children.

src/main/scala/RewriteView.scala
rachitnigam: OMG it was an actual TAB. I've never seen one before. Idk how it happened.

PR #75, opened 2019-02-19
While loops

### LINKED ISSUE #63: While loop support
Can we support dumb while loops that don’t support any unrolling?

@sa2257 what does HLS do with while loops?

rachitnigam:
Fixes #63


sampsyo:
Looks good. To state the obvious, while loops cannot be unrolled in this version, so we don't need any reasoning about loop control/loop-carried dependencies.

rachitnigam:
Yeah, that's the idea!

PR #81, opened 2019-02-21
Implement else branches

### LINKED ISSUE #76: Else statement
I keep finding myself writing this pattern:
```
if (cond) {
...
} 

if (!cond) {
...
}
```
Doesn't seem critical but would make my soul feel better if there was an `else` construct.

rachitnigam:
Fixes #76.

The type checker implements mostly we talked about in the meeting. The one new behavior is that both branches are required to use expressions that acquire capabilities in the same way. One branch cannot write into an array access while the other tries to read from it.

Also, this PR restructures `Environment` to better hide implementation details and even adds comments for it's methods!

### DIFF
--- a/src/main/scala/Emit.scala
+++ b/src/main/scala/Emit.scala
@@ -73,7 +73,8 @@ private class Emit extends PrettyPrinter {
     case CPar(c1, c2) => c1 <@> c2
     case CSeq(c1, c2) => c1 <@> text("//---") <@> c2
     case CLet(id, typ, e) => typ.get <+> value(id) <+> equal <+> e <> semi
-    case CIf(cond, cons) => "if" <> parens(cond) <> scope (cons)
+    case CIf(cond, cons, alt) =>
+      "if" <> parens(cond) <> scope (cons) <+> "else" <> scope(alt)
     case CFor(range, par, combine) =>
       "for" <> parens {
         "int" <+> range.iter <+> "=" <+> value(range.s) <> semi <+>
--- a/src/main/scala/Errors.scala
+++ b/src/main/scala/Errors.scala
@@ -2,6 +2,7 @@ package fuselang
 
 object Errors {
   import Syntax._
+  import TypeEnv.Capability
   import scala.util.parsing.input.Position
 
   def withPos(s: String, pos: Option[Position]) =
@@ -39,7 +40,7 @@ object Errors {
     s"Bank $bank in dimension $dim of $id already consumed.", None)
 
   // Invalid Capability error
-  case class InvalidCap(expr: Expr, exp: String, actual: String) extends TypeError(
+  case class InvalidCap(expr: Expr, exp: Capability, actual: Capability) extends TypeError(
     s"This expression requires $exp capability, but previous usage inferred $actual capability.", Some(expr.pos))
 
   case class AlreadyWrite(e: Expr) extends TypeError(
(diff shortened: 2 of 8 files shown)


sampsyo:
To further explore the reasoning behind that last bit of complexity, which rules out this test program:

```
decl a: bit<10>[2 bank 2];
if (true) {
  a[0] := 1;
} else {
  let x = a[0];
}
```

It seems like the underlying reason for this restriction is due to "overzealous CSE," which might allow reads or writes to `a[0]` _after_ the `if`. Of course, neither should be allowed—the bank _might_ have been read _or_ written, which should preclude _both_ reads and writes that follow.

However, wouldn't this be safe if the capabilities were "constrained" to the scopes where they're manufactured, on either branch of the `if`? As long as they don't leak out of the `if`, but the `a[0]` bank is still unavailable, this should be safe—unless I'm missing something!

rachitnigam:
The distinction ends up matter because of we haven’t decided if we want hardware enables or not. If we don’t want enables, then all reads and writes need to be muxed (missing read or write is just a noop).

If you truly disable the hardware then it won’t matter. Either way, removing the restriction isn’t hard at the type checking level.

rachitnigam:
Hm, another thing to consider is that this capability check doesn't correctly work for functions. If you call `f()` and `g()` in one branch each and `f` reads to a bank while `g` writes to it, then the capability check will not exclude the program. It will still not allow programs that use the same array in parallel composition with the `if-else`.

This is a good enough reason to remove the check.

Tangential: Functions already don't have a macro respecting semantics. A call conservatively consumes all the banks in an array, whereas inlining the expression might not completely consume it. However, I am fine with that for now since it's a conservative choice.

sampsyo:
> This is a good enough reason to remove the check.

Cool, yeah, I agree. This is a somewhat tricky topic, but I actually think there's an argument to be made that the above program (without functions) is safe, even without a semantics that exposes whether the implementation strategy is "just muxing" or includes an enable. It's certainly safe in the "enable" world. In the "just muxing" world, it seems like this could be implemented with enough muxing such that only the read or write "message" actually gets "sent" to the memory.

Here's a somewhat illegible diagram showing what I mean:
read-or-write

This abstract memory has two "input ports," the address (used for both reads and writes) and the address (used only for writes). In reality, there is one more bit here: the single input flag that tells the memory whether we're doing a read or a write. The address needs to have muxing to get its values from the `then` or `else` branch, just as it would need to if the program just had two reads instead of a read and a write. To make a read+write condition safe, we just _also_ need a mux on the address (and that extra read/write bit) to make sure only one instruction gets "sent" to the memory.

> Tangential: Functions already don't have a macro respecting semantics. A call conservatively consumes all the banks in an array, whereas inlining the expression might not completely consume it. However, I am fine with that for now since it's a conservative choice.

Indeed! We should probably think of this as the essential difference between our functions and ordinary macros: our functions support intraprocedural (and therefore compositional) reasoning.

rachitnigam:
That's a neat diagram! I understand your point about disables. It seems there are several ways of "disabling" the input/output signals for read and writes. For example in this one, we run the hardware by remove the signal completely.

Figuring out which one of these disables are actually the "right ones" for fuse2rtl will be useful. It could also be the case that this analysis is too complicated and we truly need to have enables to make if else work correctly.

I'll merge after I remove this example. LMK if you want other changes.

sampsyo:
Exactly, that's a good point—I'd be interested to see what turns out to be the most straightforward implementation strategy.

rachitnigam:
Well, it's already implemented here.

sampsyo:
Oops, sorry—I meant the fuse2rtl implementation strategy.

PR #83, opened 2019-02-21
Struct support

### LINKED ISSUE #61: Add `struct` support
For some of the MachSuite apps, it would be helpful to have some kind of "and" type, like a `struct`, that could be declared in the Fuse body (by Fuse body I mean not declared with `decl`, which would get compiled to a function argument). Would be helpful in particular in:
- `bfs/queue`: need a way to represent a node of a graph.
- `md/grid`: in the C code, has vectors represented as `struct`s, that are declared in the "Fuse body" I referred to above. I thought of just using Fuse arrays to represent the vectors, but they can only be declared with `decl`.

rachitnigam:
Fixes #61.

- [x] Restrict usage of arrays in structs
- [x] Allow writing struct literals.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,6 +17,7 @@ scalacOptions ++= Seq(
   "-Ywarn-unused",
   "-Ywarn-value-discard",
   "-Yno-adapted-args",
+  "-Xfatal-warnings"
 )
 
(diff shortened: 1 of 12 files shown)


rachitnigam:
Ready to merge.

sampsyo:
Yay!! LGTM.

PR #84, opened 2019-02-22
Monadic rewrite for RewriteViews

rachitnigam:
Addresses https://github.com/cucapra/seashell/pull/71#discussion_r257100246

### DIFF
--- a/src/main/scala/RewriteView.scala
+++ b/src/main/scala/RewriteView.scala
@@ -1,5 +1,8 @@
 package fuselang
 
+import Syntax._
+import CodeGenHelpers._
+
 /**
  * AST pass to rewrite views into simple array accesses. Should be used after
  * type checking.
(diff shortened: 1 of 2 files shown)


src/main/scala/RewriteView.scala
```diff
@@ -15,103 +18,82 @@ package fuselang
...
-    case eaa@EArrAccess(arrId, idxs) => {
-      val (idxsn, env1) = foldExprs(idxs)
-      // If the array id is a view array rewrite it and recur
+  def rewriteExpr(e: Expr): State[T, Expr] = e match {
+    case EVar(_) | EInt(_, _) | EFloat(_) | EBool(_) | _:ERecAccess => State.unit(e)
+    case eb@EBinop(_, e1, e2) => for {
+      e1n <- rewriteExpr(e1)
+      e2n <- rewriteExpr(e2)
+    } yield eb.copy(e1 = e1n, e2 = e2n)
+    case eaa@EArrAccess(arrId, idxs) => State { env =>
```
rachitnigam: @sampsyo Is there a nice monadic function to represent what I'm doing here? This case requires access to the state (`env`) and the expression (`idxns`) to work.

sampsyo: Hmm; I don't think so! I think this might be the simplest way to write this, AFAICT.

src/main/scala/RewriteView.scala
```diff
@@ -15,103 +18,82 @@ package fuselang
  * If `a` itself is a view, we keep rewriting it until we reach a true array.
  */
 object RewriteView {
-  import Syntax._
-  import CodeGenHelpers._
+  import Utils.State
 
-  // We can use a flat environment because the type checker makes sure things
-  // in the right scope.
   type T = Map[Id, List[Expr] => Expr]
```
sampsyo: A tiny thing, but any particular reason to call this `T` instead of `Env` to suggest its purpose?

rachitnigam: Yeah, mostly because I wanted to extract this into a generic class.

rachitnigam: Is this a pattern that can be better expressed with arrows? (Not saying I’ll rewrite it into arrows)

sampsyo: My apologies; I just realized I left this question hanging: I don't currently see a great way to use arrows here, but I freely admit to not being any kind of typeclass wizard. :smiley: A real Haskell person may be able to see exactly how to do it.

rachitnigam: Turns out there is: https://github.com/cucapra/seashell/blob/985d5d35ea97650b29fe17a2e7d17d2d475c6d4a/src/main/scala/RewriteView.scala#L38

Only took me two hours to figure out the types!

sampsyo: Very nicely done!

PR #85, opened 2019-02-27
Docusaurus documentation

rachitnigam:
Already live here: https://cucapra.github.io/seashell/


sampsyo:
Cool! Docusaurus is pretty nice. Here are a few comments:

- I personally don't love GitHub pages as a deployment option. Can you try using Hooknook to deploy on our own servers?
- Does using Docusaurus require us to commit copies of all this boilerplate under the `website/` directory? All of that (including the README) looks like it's unmodified stuff from Docusaurus itself, which is pretty crufty. Other documentation generators I've used (e.g., [Sphinx](http://sphinx-doc.org) or [mdbook](https://github.com/rust-lang-nursery/mdBook)) work by reading "just" the source files in Markdown or whatever along with a simple configuration.
- This PR also removes some design notes we've written before that I still think are relevant but are not exactly user-facing documentation. @rachitnigam, I'll leave it to you: do you think those should get wrapped up in the new documentation infrastructure, or should we leave them in a separate place because of their non-documentation nature?

rachitnigam:
1. Yeah, I can look at a hooknook deployment. Since docusaurus supports building a full website, the code under website can be customized for that in the future.

2. The old docs have been moved to notes/. I didn’t intend on deleting them at all. Seems like they are still live under the Capra domain.

sampsyo:
Oh oops! Fooled again by git's diffstat output showing `D` for moved files…

rachitnigam:
Website available at: https://capra.cs.cornell.edu/fuse

Issue #70, opened 2019-02-14
Array access using dynamic ints


rachitnigam:
Some of the examples in the Xilinx manual allow for this:
```
void inp(int a[10], int readline) {
  int c = a[readline/2];
}
```
It seems we can conservatively allow for these by making dynamic accesses consume all banks of the array.

sampsyo:
It occurs to me that allowing dynamic access requires all-way muxing, i.e., if the array is banked, we need a view that is unbanked (unibanked). Which is the same hardware implementation and typing effect as for our `flex` view proposal (c.f. #57) in the special case where the result has only a single bank.

So the policy should likely be that unbanked arrays can be accessed with dynamic integers. (And you can use `flex` to get an unbanked array view if you need to.)

rachitnigam:
This sounds reasonable. Concretely, the `consumeBanks` function in `typeCheck` needs to change to allow for `TSizedInt` and fully consume the dimension. The case statement should also make sure that the dimension is unbanked and reject the access if it's not.

This can be done in parallel with the `flex` proposal.

PR #88, opened 2019-02-27
adding TSizedInt rule for consumeBanks

apurvakoti:
Addresses issue #70 . 
Checks if the array is unbanked (by checking if its banking factor is 1), and if it is, consumes bank 0.

Not sure about the error to throw here, or if bres * 1 is the correct mapping here.

### DIFF
--- a/src/main/scala/Errors.scala
+++ b/src/main/scala/Errors.scala
@@ -42,6 +42,9 @@ object Errors {
   case class AlreadyConsumed(id: Id, dim: Int, bank: Int) extends TypeError(
     s"Bank $bank in dimension $dim of $id already consumed.")
 
+  case class InvalidDynamicIndex(id:Id, bf:Int) extends TypeError(
+    s"Dynamic access of array $id requires unbanked array. Actual banking factor: $bf. Use a shrink view to create unbanked array.", id.pos)
+
   // Invalid Capability error
   case class InvalidCap(expr: Expr, exp: Capability, actual: Capability) extends TypeError(
     s"This expression requires $exp capability, but previous usage inferred $actual capability.", expr.pos)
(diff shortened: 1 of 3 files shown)


rachitnigam:
rebase with master please.

src/main/scala/TypeCheck.scala
```diff
@@ -63,6 +63,9 @@ object TypeChecker {
           env2.update(id -> env2(id).consumeDim(i, e - s)) -> bres * (e - s)
         case (TStaticInt(v), env2) =>
           env2.update(id -> env(id).consumeBank(i, v % dims(i)._2)) -> bres * 1
+        case (TSizedInt(_), env2) =>
+          if (dims(i)._2 != 1) throw BankUnrollInvalid(dims(i)._2, 1) //different error?
```
rachitnigam: Add a new error called `InvalidDynamicIndex` based on the `InvalidIndex` error message and explain the conditions required to index with a dynamic int.

rachitnigam:
Logic looks good. Add the new error message and tests in the `TypeCheckerSpec.scala` under the "Array access" tests.

src/main/scala/Errors.scala
```diff
@@ -42,6 +42,9 @@ object Errors {
   case class AlreadyConsumed(id: Id, dim: Int, bank: Int) extends TypeError(
     s"Bank $bank in dimension $dim of $id already consumed.")
 
+  case class InvalidDynamicIndex(id:Id) extends TypeError(
+    s"Cannot dynamically access unbanked array $id.", id.pos)
```
rachitnigam: "Dynamic access of array $id requires unbanked array. Actual banking factor: $bf. Use a shrink view to create unbanked array"

PR #91, opened 2019-03-01
Software harness

### LINKED ISSUE #82: Software testing harness
While we're waiting for #68, I can implement a software  harness to validate that we are implementing the right functions and emitting valid C.

Implementation Sketch
----
- Write a C program (data parser) that parses a file that contains one value on each line.
- The data parser passes an array containing all the data into a function called `harness(int* data)`.
- Each Fuse program will define a harness function that will realign the data as needed. 
  - Note: We probably need explicit casting for this.
- Compile benchmark with the data parser (using gcc) and run it on a supplied data file.
- Add a parameter to the CLI to turn all sized ints into either `double` or `int`. When compiling for the harness, use this parameter.

rachitnigam:
Fixes #82.

~WIP.~

This PR defines a new backend and a C++ parsing harness that will allow us to run kernels with arbitrary data.


sampsyo:
Cool. Here are a couple of things I notice in this version:

1. Looks like you have a new abstract concept of a backend target, which lets you customize (for example) the strings used as type names for various internal types. This seems useful and, as you alluded to yesterday, could be the right groundwork to lay to support other HLS tools as targets. (Like Merlin, for example.)
    - One small thing that could cause problems: sized ints in the "Vivado-like" backend are emitted as C `int`s, which will of course not quite match the semantics of sized numbers. They will overflow when they shouldn't for 200-bit ints, for example, and they will *not* wrap around when they should for 4-bit ints. As a purely "quality of life" issue, it would be nice if the default backend for software execution faithfully implemented the source language semantics for sized numbers. This would be really heavyweight, but we could link to something like [gmp](https://gmplib.org) to provide arbitrary-precision ints.
2. As a separate-ish new feature, the compiler can now emit a `main` function including a bunch of parsing logic that lets you actually *run* the programs.
     - I have to admit that I'm fundamentally a little bit wary of the approach that involves *generating* the main file. In general, real programs—and perhaps even our own examples!—will want to do different, squirrelly things as part of `main`. For example, you could imagine that some benchmark might be best run in some outer loop that iterates the kernel to convergence. It also seems a little odd for the compiler to include code for parsing a special data format, when "real" users of our system (not our little examples!) will of course have their own data formats, so they'll never want to use that part.
     - That's why I've been recommending the approach where each of our examples comes with its own C/C++ file that has `main` in it. To avoid duplication, they can all share a common header with your file-parsing code, but I think it might be cleaner for the per-application `main` to be "handcrafted," one by one, rather than automatically generated.

rachitnigam:
Ah, the PR name is confusing. This is NOT meant for benchmarking.

This is something that I’d imagine using when developing a kernel and wanting some nice REPL like interaction by doing fuse prog run. This approach definitely doesn’t scale to all benchmarks (say BFS).

The benchmark related parts is the concept of a new backend. I can create a separate backend that understands the software harness setup after this PR.

Regarding bit accuracy, I’ve been looking into GMP. The abstraction might make it easier to link with it regardless of which software backend we’re using.

The final benefit of this backend is being able to test the emitted code is valid CPP.

sampsyo:
Aha, cool! So I should think of the purpose of the `main` emitter stuff as being like an alternative to having an interpreter (with the side effect of quickly validating that the codegen produces valid C++). Sounds good!

sampsyo:
An alternative to gmp that was just discussed at the meeting I'm at right now at UCLA is Mentor Graphics' [AC Datatypes](https://hlslibs.org) library, which has [header-only C++ libraries](https://github.com/hlslibs/ac_types/tree/master/include) called `ac_int`, `ac_fixed`, and `ac_float` implementing signed numbers. Could be a little lighter-weight than full gmp.

rachitnigam:
Ready for review!

High level summary:
- Changes to CLI:
  - `-o` put the output from the compiler in a file
  - `-b` switch to a different backend. Supported backends are `vivado` and `c++`
  - `fuse run <src> -o <out>` generate an executable that can be run with a json data file.
- To review, try to follow the example added under `docs/cpp-runnable.md`.
  - `sbt assembly` automatically downloads the required headers.
- Take a look at the output of `matadd.cpp` after following the previous step. Make sure the emitted code makes sense. 
- Take a look at the `parser.cpp` file.

I'll try to merge this tomorrow before the meeting. Everyone should try playing with this PR. Requiring review from @sa2257 and @sampsyo.

src/main/scala/GenerateExec.scala
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Provides utilities to compile a program and link it with headers required
+ * by the CppRunnable backend.
+ */
+object GenerateExec {
+  val headers = List("parser.cpp", "picojson.h")
+
+  var headerLocation = Paths.get("src/main/resources/headers")
+  val headerFallbackLocation = Paths.get("_headers/")
```
rachitnigam: If the command is run from outside the compiler directory, `fuse run` will extract the required headers into a `_headers/` folder as a fallback.

docs/cli-compile.md
```diff
@@ -11,3 +11,22 @@ It provides the following options:
...
+The `-b` or `--backend` allows changing the backend used for generating code.
+Support backends are:
+
+1. `vivado`: Generates code compatible with Vivado HLS.
+2. `c++`: Generate code that can be compiled into an executable. See [this](cpp-runnable.md).
+
+## `run`
+
+The subcommand `run` requires the `-o` to be present. It automatically generates
+a CPP file and links it with the json data parser. See [this](cpp-runnable.md)
```
sampsyo: More silly writing nitpicking: C++, not CPP; and JSON is capitalized. Finally, "this" is rarely all that helpful as text for a link—I recommend:

    See ["Generating Executables"](cpp-runnable.md) for details.

Also, it occurs to me that the default mode also generates a C++ file. So this might not be the clearest way to describe what the purpose of this mode is. Maybe a better way to introduce it would be to say that this is a mode for testing of these programs that lets you build a "standalone" executable by including a `main` function that lets you run on JSON data files.

docs/cli-compile.md
```diff
@@ -11,3 +11,22 @@ It provides the following options:
...
+for more information.
```
sampsyo: This is just an aesthetic UI argument, but it's weird that the CLI has an *optional* subcommand. That is, most CLIs I know either have subcommands or don't—`git` has them, obviously, and there's no way to *just* invoke `git` without a subcommand. Same with `svn` before it.

For whatever reason, this feels to me like it would be just fine as a flag. That is, `fuse --run source.file` seems to work just fine and avoids the "optional subcommand" thing.

docs/cpp-runnable.md
```diff
@@ -0,0 +1,85 @@
+---
+title: Generating Executables
+id: cpp-runnable
+---
+
+The `fuse` CLI provides the `run` subcommand for generating executables that
+can be used to test Fuse code without going throught a synthesis workflow.
```
sampsyo: More saliently, I'd say it's useful for testing Fuse code without writing your own host-side wrapper code (i.e., `main` function).

docs/cpp-runnable.md
```diff
@@ -0,0 +1,85 @@
...
+
+To build an executable, first run the following command from the project root:
+
+```
+sbt assembly
+```
+
+This will rebuild the compiler and download any headers required for building
+the executable.
```
sampsyo: Everyone will have already done that in order to use the compiler at all, right?

docs/cpp-runnable.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+fuse run matadd.fuse -o matadd.cpp
+```
+
+the command will something similar to the following:
+
+```sh
+g++ --std=c++11 -Wall -Isrc/main/resources/headers matadd.cpp -o matadd.cpp.o
+```
```
sampsyo: This wording seems to suggest that `fuse run` *just* executes `g++`. But that's not quite true: it first compiles the source to C++ (including the wrapper) and then compiles it the rest of the way. I'd just say that, describing all the things the command does—and maybe even omit this command, the precise details of which are maybe not terribly relevant.

src/main/resources/headers/parser.cpp
```diff
@@ -0,0 +1,141 @@
+#include<iostream>
+#include<fstream>
+#include<vector>
+#include<string>
```
sampsyo: Missing spaces look a little weird?

src/main/scala/backends/CppLike.scala
```diff
@@ -0,0 +1,148 @@
...
+      case EApp(fn, args) => fn <> parens(hsep(args.map(emitExpr), comma))
+      case EInt(v, base) => value(emitBaseInt(v, base))
+      case EFloat(f) => value(f)
+      case EBool(b) => value(if(b) 1 else 0)
+      case EVar(id) => value(id)
+      case EBinop(op, e1, e2) => parens(e1 <+> op.toString <+> e2)
+      case EArrAccess(id, idxs) => id.typ match {
+        case Some(TArray(_, dims)) => id <> brackets(flattenIdx(idxs, dims.map(_._1)))
+        case Some(_) => throw Impossible("array access doesnt use array")
+        case None => throw Impossible("type information missins in exprToDoc")
```
sampsyo: typo: missins

docs/cpp-runnable.md
rachitnigam: #114 fixed this.

src/main/resources/headers/parser.cpp
rachitnigam: Fixed.

src/main/scala/backends/CppLike.scala
rachitnigam: Fixed.

src/main/resources/headers/parser.cpp
rachitnigam: Fixed.

build.sbt
```diff
@@ -25,3 +25,26 @@ parallelExecution in Test := false
...
+  import java.io.File
+  import java.net.URL
+
+  val picoJsonHdrLoc = new File("src/main/resources/headers/picojson.h")
+
+  if (!picoJsonHdrLoc.exists()) {
+    val picoJsonHdr = new URL("https://raw.githubusercontent.com/kazuho/picojson/master/picojson.h")
+    val cmd = s"wget $picoJsonHdr --directory-prefix=${picoJsonHdrLoc.toString}"
+    // sys.process DSL magic!
+    picoJsonHdr #> picoJsonHdrLoc !!
```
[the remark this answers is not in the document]
rachitnigam: Fixed.

src/main/scala/GenerateExec.scala
```diff
@@ -0,0 +1,60 @@
...
+    // Make sure all headers are downloaded.
+    for (header <- headers) {
+      if (Files.exists(headerLocation.resolve(header)) == false) {
+        throw HeaderMissing(header, headerLocation.toString)
+      }
+    }
+
+    // Generate [[out]]. `!` is defined by sys.process:
+    // https://www.scala-lang.org/api/2.12.8/scala/sys/process/index.html
+    val cmd = s"$CXX ${src.toString} -o $out"
```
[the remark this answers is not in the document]
rachitnigam: Fixed.

docs/cli-compile.md
rachitnigam: So “run” implies a bunch of other things such as backend=c++ and the fact that -o is required.

I tried making —run a flag but the code for enforcing the above constraints is non local and annoying to read. For example, the code has to say if mode=run then validate that -o is not None and backend=c++. Note that having a subcommand doesn’t stop us from specific common things like kernelName

For uniformity’s sake, I can make a fuse compile command. Would that work?

rachitnigam:
Merging now. @sa2257 will review and ask questions here by EOW.

PR #100, opened 2019-03-01
[WIP] MachSuite SPMV

tedbauer:
[Link to original MachSuite app.](https://github.com/breagen/MachSuite/blob/master/spmv/crs/spmv.c)

## Compilation Status
Compiles!!!

## Issues
1. This app has a loop with a dynamic start/end, like BFS (#94). I wrote it with a `while` loop, so it can't be unrolled. They unroll by a factor of 64 in the original app.

## Todo
- [ ] Address magic numbers
- [ ] Optimize with `bank` and `unroll`(?)


tedbauer:
To be specific about (1), here's the original program:
```
    ...
    spmv_1 : for(i = 0; i < N; i++){
        sum = 0; Si = 0;
        int tmp_begin = rowDelimiters[i];
        int tmp_end = rowDelimiters[i+1];
        spmv_2 : for (j = tmp_begin; j < tmp_end; j++){
            Si = val[j] * vec[cols[j]];
            sum = sum + Si;
        }
        out[i] = sum;
    }
}
```
`spmv_1` and `spmv_2` are each unrolled by 8, and each array (`val`, `vec`, `cols`, `out`) has 64 banks. With the currently pushed Fuse program, I can't see how to get any parallelization.

PR #102, opened 2019-03-01
Subtyping is hard

### LINKED ISSUE #53: Allow specification of custom default bit lengths
@sampsyo said the following in #46:
> It would be nice to make sure that the number fits into a 32-bit size. And even if not, it could be sensible to put the default integer type into a constant somewhere.

> I realize this seems like a low-level detail, but this could actually be a deeper question: we might want to let people _specify_ the default numeric type within a scope. Because numeric widths are so important for accelerator efficiency, people often want to decide the numeric type for an entire design or module at once. That should probably affect (most?) literals as well as the buffers they eventually flow into. Seems sort of tricky.

This can be implemented in two ways: (1) Parsing magic comments in fuse files that specify bit lengths, or (2) Making default bit length a CLI option.

### LINKED ISSUE #87: Think hard about resolve types
Having typedefs in the language is non-trivial. I need to think hard about type defs and if they work correctly.

### LINKED ISSUE #95: `==` between static and non-static things
In the BFS app (see #94), this code:
```
if (tmp_l == horizon) {...}
```
throws this error:

```
[Type error] [31.15] Expected subtype of static(0) in ==, received: idx((0,1), (0,50)).
    if (tmp_l == horizon) {
              ^
```

Here's the whole program, for convenience:
```
record edge_t {
  dst: bit<32>;
};

record node_t {
  edge_begin: bit<32>;
  edge_end: bit<32>;
};

decl nodes: node_t[50];
decl edges: edge_t[50];
decl starting_node: bit<32>;
decl level: bit<32>[50];
decl level_counts: bit<32>[50];

let cnt = 0;

// Temporary variables for serial updates
let tmp_begin = 0;
let tmp_end   = 0;
let tmp_dst   = 0;
let tmp_level = 0;

let tmp_l = 0;

for (let horizon = 0..50) {
  cnt := 0;
  for (let n = 0..50) {
    tmp_l := level[n];
    ---
    if (tmp_l == horizon) {
      tmp_begin := nodes[n].edge_begin;
      ---
      tmp_end := nodes[n].edge_end;
      ---
      for (let e = 0..50) { // FIXME: find upper bound
        if (e >= tmp_begin && e <= tmp_end) {
          tmp_dst := edges[e].dst;
	  tmp_level := level[tmp_dst];
	  ---
	  if (tmp_level == MAX_LEVEL) {
            level[tmp_dst] := horizon + 1;
	    ---
	    cnt := cnt + 1;
	  }
	}
      }
    }
  }
}
```

rachitnigam:
Fixes #53. Fixes #95. Fixes #87.

Three important things:
- Explicit subtyping hierarchy: sized int > idx types > static ints
- `let x = 1` `x` gets the type `TSizedInt(bitsNeeded(1))`.
- Only numbers and binary operations on numbers can a static type. So `a[0]` still works fine, but `let x = 1; a[x];` doesn't.

### DIFF
--- a/src/main/scala/Subtyping.scala
+++ b/src/main/scala/Subtyping.scala
@@ -2,7 +2,6 @@ package fuselang
 
 import scala.math.{max,log10,ceil}
 import Syntax._
-import Errors.NoJoin
 
 /**
  * Subtyping relations are only defined over the number hierarchy. Read 't1 < t2'
@@ -47,6 +46,7 @@ object Subtyping {
   def areEqual(t1: Type, t2: Type) = (t1, t2) match {
     case (TStaticInt(v1), TStaticInt(v2)) => v1 == v2
     case (_:TIndex, _:TIndex) => true
+    case (_:TFloat, _:TFloat) => true
     case _ => t1 == t2
   }
 
@@ -61,26 +61,26 @@ object Subtyping {
     case _ => areEqual(sub, sup)
   }
 
-  def hasJoin(t1: Type, t2: Type): Boolean = (t1, t2) match {
-    case (_:IntType, _:IntType) => true
-    case _ => false
-  }
-
-  def joinOf(t1: Type, t2: Type, op: (Int, Int) => Int) = (t1, t2) match {
-    case (TStaticInt(v1), TStaticInt(v2)) => TStaticInt(op(v1, v2))
-    case (TSizedInt(s1), TSizedInt(s2)) => TSizedInt(max(s1, s2))
+  def joinOf(t1: Type, t2: Type, op: BOp): Option[Type] = (t1, t2) match {
+    case (TStaticInt(v1), TStaticInt(v2)) => op.toFun match {
+      case Some(fun) => Some(TStaticInt(fun(v1, v2)))
+      case None => Some(TSizedInt(max(bitsNeeded(v1), bitsNeeded(v2))))
+    }
+    case (TSizedInt(s1), TSizedInt(s2)) =>
+      Some(TSizedInt(max(s1, s2)))
     case (TStaticInt(v), TSizedInt(s)) => {
-      TSizedInt(max(s, bitsNeeded(v)))
+      Some(TSizedInt(max(s, bitsNeeded(v))))
     }
     case (TSizedInt(s), TStaticInt(v)) => {
-      TSizedInt(max(s, bitsNeeded(v)))
+      Some(TSizedInt(max(s, bitsNeeded(v))))
     }
     case (idx@TIndex(_, _), TStaticInt(v)) =>
-      TSizedInt(max(idx.maxVal, bitsNeeded(v)))
+      Some(TSizedInt(max(idx.maxVal, bitsNeeded(v))))
     case (TStaticInt(v), idx@TIndex(_, _)) =>
-      TSizedInt(max(idx.maxVal, bitsNeeded(v)))
-    case (_: TIndex, t2@TSizedInt(_)) => t2
-    case (t2@TSizedInt(_), _:TIndex) => t2
-    case (t1, t2) => throw NoJoin(t1, t2)
+      Some(TSizedInt(max(idx.maxVal, bitsNeeded(v))))
+    case (_: TIndex, t2@TSizedInt(_)) => Some(t2)
+    case (t2@TSizedInt(_), _:TIndex) => Some(t2)
+    case (_:TFloat, _:TFloat) => Some(TFloat())
+    case _ => None
   }
 }
--- a/src/main/scala/Subtyping.scala
+++ b/src/main/scala/Subtyping.scala
@@ -1,0 +1,51 @@
+package fuselang
+
+import scala.math.{max,log10,ceil}
+import Syntax._
+import Errors.NoJoin
+
+/**
+ * Subtyping relations are only defined over the number hierarchy. Read 't1 < t2'
+ * as t1 is subtype of t2. The subtyping hierarchy is:
+ *
+ * sized(n) > sized(n - 1) ... > sized(1) > idx > static
+ *
+ * Note that all idx types and static types are equal to each other.
+ */
+object Subtyping {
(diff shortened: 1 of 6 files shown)


src/main/scala/Subtyping.scala
```diff
@@ -0,0 +1,51 @@
...
+ * Note that all idx types and static types are equal to each other.
+ */
+object Subtyping {
+  def bitsNeeded(n: Int) = n match {
+    case 0 => 1
+    case n => ceil(log10(n)/log10(2)).toInt + 1
+  }
+
+  def areEqual(t1: Type, t2: Type) = (t1, t2) match {
+    case (_:TStaticInt, _:TStaticInt) => true
```
sampsyo: Perhaps this should be:

```scala
case (_:TStaticInt(v1), _:TStaticInt(v2)) => v1 == v2
```

i.e., static int types are equal when the numbers are equal (and unordered otherwise).

src/main/scala/Subtyping.scala
```diff
@@ -0,0 +1,51 @@
...
+    case (_:TIndex, _:TIndex) => true
+    case _ => t1 == t2
+  }
+  def isSubtype(sub: Type, sup: Type): Boolean = (sub, sup) match {
+    case (TSizedInt(v1), TSizedInt(v2)) => v1 <= v2
+    case (_:IntType, _:TSizedInt) => true
+    case (_:TStaticInt, _:TIndex) => true
+    case (TArray(tsub, _), TArray(tsup, _)) => isSubtype(tsup, tsub)
```
sampsyo: Because arrays are mutable, it's probably safest to require invariant subtyping for them—i.e., either `areEqual(tsup, tsub)` or `isSubtype(tsup, tsub) && isSubtype(tsub, tsup)`. Otherwise, you could imagine something silly like this:

```
def foo(b: bit<1000>[20]) {
  b[0] := GIANT_NUMBER;
}

decl a : bit<10>[20];
foo(a);
```

i.e., to let an alias to the memory "disagree" about the right type of numbers the memory should contain.

rachitnigam:
Some discussion about from today's meeting. @sa2257 and Kenneth pushed back against the idea of automatically inferring the the max bits needed in a static bindings. So in this PR, we have:

`let x = 1` where type of `x` is `bit<1>` while `let x: bit<32> = 1` assigns `x` type to be `bit<32>`. 

However, this might not seems very intuitive. We can three alternative implementations:

1. `let x = 1` where type of `x` is `bit<32>` always. It mirrors C's default.
2. Have a `--bit-width` flag that defines a default for the program and emitter and make the default width 32.
3. Have scope level control on bit widths. So something like:
```
{bw = 2
  let x = 1 // x: bit<2>
}
{bw = 4
  let y = 1 // y: bit<4>
}
```

Finally, polymorphism and type level constants might just make these things useless. For example, we would be able to have:
```
type BW: Nat = 10
let x: bit<BW> = 1 // x: bit<10>
let y = 1 // y: bit<10>
```

sampsyo:
Perhaps @sa2257 and Kenneth (Kenneth, are you on GitHub? :smiley:) can elaborate on what is unintuitive about the fact @rachitnigam mentioned above? That is, it's always possible to specify all your bit widths explicitly—is the problem just that you might forget to do that?

I'd prefer my old proposal about building on type aliases (https://github.com/cucapra/seashell/issues/53#issuecomment-462168448) over a special-purpose language feature like `bw=` scopes and compiler flags, which are a little ad hoc.

src/main/scala/Subtyping.scala
rachitnigam: That'll disallow `1 == 2` since `1 : static(1)` and `2 : static(2)`.

I think I might be conflating subtyping and joins. In this case, we really just want a join on the two static ints. Any pointers on which one should be used where?

sampsyo: Good question. One way of looking at is imagining that you have a polymorphic function `equal[T](lhs: T, rhs: T)`. Then you have a call to this function without a type parameter, like `equal(1, 2)`, where the type `T` needs to be inferred. The answer to the question, "is there a type `T` that makes this call well-typed?" is equivalent to "do the types of `1` and `2` have a least upper bound, i.e., a join?"

For this particular case, following this philosophy would say "I can compare 1 and 2 because I can subsume them both to `bit<2>` and do the comparison as I would on `bit<2>` values." This would also nicely handle comparisons between static and non-static numbers.

src/main/scala/Subtyping.scala
rachitnigam: Fixed.

rachitnigam:
Added more subtyping fixes.

src/main/scala/Subtyping.scala
```diff
@@ -0,0 +1,86 @@
...
+    case (_:TFloat, _:TFloat) => true
+    case _ => t1 == t2
+  }
+
+  def isSubtype(sub: Type, sup: Type): Boolean = (sub, sup) match {
+    case (TSizedInt(v1), TSizedInt(v2)) => v1 <= v2
+    case (_:IntType, _:TSizedInt) => true
+    case (_:TStaticInt, _:TIndex) => true
+    case (TArray(tsub, _), TArray(tsup, _)) => {
+      // Arrays are mutable so we are conservative and disallow subtyping.
```
sampsyo: Or, to be perfectly pedantic, array subtyping is invariant. More on this in 6110 soon! :smiley:

sa2257:
My concerns are over use of automatic inference and on making them default. I added some comments to #53

rachitnigam:
@sampsyo @sa2257 ping. Is the discussion in #53 satisfactory?

sampsyo:
Yeah, this is ready to merge IMO. We can keep #53 open to track the longer-term discussion about the type inference behavior in `let x = 5`.

rachitnigam:
Cool, the docs reflect the behavior w.r.t inference: https://capra.cs.cornell.edu/fuse/docs/types/#sized-integers.

PR #107, opened 2019-03-03
buildbot: Simplify FPGA execution stage

sampsyo:
These are several changes to the buildbot worker stage where we execute code on the ZedBoard. It includes cleaning up unused code and clarifying what was left.

One outstanding issue I mentioned in a commit message: it seems like 120 seconds is probably overkill for the delay where we wait for the ZedBoard to restart. I'm guessing something more like 10 seconds would be fine, but I'll need to try it out.

I haven't yet had a chance to test this on Gorgonzola. I'll try to do that soon (and then merge), but @sa2257, you're welcome to do that too.


sa2257:
@sampsyo sleep time is probably an overkill. I didn't bother changing that since the bottleneck is the synthesis time which takes 15-30 minutes.

sampsyo:
OK; thanks! I changed the time to 40 seconds, which I think should be more than enough.

sampsyo:
Tested & fixed some bugs. Seems to work!
http://gorgonzola.cs.cornell.edu:8000/jobs/R5FELgVnjtY.html

Issue #105, opened 2019-03-01
Low Priority Issue Tracker


tedbauer:
This issue tracks low-priority compiler issues that aren't important enough to warrant their own issue.

## Tasks

- [x] Semicolon parsing
- [ ] Program structure error messages
- [ ] Fix https://github.com/cucapra/seashell/issues/105#issuecomment-468972729
- [x] Nonexistent file error
- [x] Record syntax
- [ ] Fixed sized arithmetic for C++ backend
- [x] `double` type support

### ~~1. Semicolon Parsing~~
The compiler parses loops/conditionals like this:
```
for (..) {
    ...
};
let x = 2;
...
```
It would be nice to not require a `;` after the`}` closing the loop body.

### 2. Program structure error messages
This program:
```
let x = 2;
decl test: bit<32>[5];
```
Gives this error message:
```
[Error] [2.6] failure: '---' expected but 't' found

decl test: bit<32>[5];
        ^
```
The error is due to the fact that `let` defs cannot occur above `decls` as Fuse program structure dictates, but this error message does not communicate that.

sa2257:
3. Assignment operator errors

```
for (let i = 0..1024) unroll 32 {
    c[i] = a[i] + b[i];
}
```

The error I get is 
```
[31mError[0m] [4.10] failure: '}' expected but '=' found

    c[i] = a[i] + b[i];
         ^
```
Would be nice to have an error that points to reassignment.

tedbauer:
4. No file error message

If you supply a nonexistent file to `fuse`, it throws some uncaught exception, would be nice if it had a prettier error message

tedbauer:
## 5. Record syntax
The docs are inconsistent with  #129 with record syntax

rachitnigam:
6. Fixed sized arithmetic support for C++ backend using [ac_int](https://hlslibs.org/#collapseACDatatypes).

tedbauer:
# 7. Add support for `double` type (?)
The type signatures in the header files for most of the MachSuite apps have `double` types, so having some type in Fuse that would emit a `double` in the C++ backend would be needed to write Fuse files that match the headers.

sampsyo:
N.B. item 7 above (double support) has landed!

# 8. Character literals

*Maybe* support literals like `'a'`? Their type would be `ubit<8>`, probably. Originally discussed in #211.

rachitnigam:
Stale and a bunch of these things are fixed.

PR #108, opened 2019-03-03
Parsing fixes

rachitnigam:
Two fixes:
- End of `if` or `for` blocks no longer require a `;`.
- Locations that require numbers to be parsed throw better error message


sampsyo:
Referencing #105, where the first bullet point was discussed.

PR #109, opened 2019-03-03
make `fuse` command insensitve to links

rachitnigam:
@sampsyo or @sa2257 can you make sure this won't break buildbot setup? Once we have this, we can freely link `fuse` globally or put it on the path.

### DIFF
--- a/fuse
+++ b/fuse
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-here=`dirname $(readlink $0) || dirname $0`
+here=`dirname $(readlink $0 || echo $0)`
 java -jar $here/target/scala-2.12/fuse.jar $@
--- a/fuse
+++ b/fuse
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-here=`dirname $0`
+here=`dirname $(readlink $0) || dirname $0`
 java -jar $here/target/scala-2.12/fuse.jar $@


sampsyo:
This is a good idea, but at least on my shell the `||` isn't quite enough to suppress the error that results when the script $0 is *not* a symbolic link:

```
$ dirname foo.sh
.
$ readlink foo.sh
$ dirname $(readlink foo.sh)
usage: dirname path
$ dirname $(readlink foo.sh) || dirname foo.sh
usage: dirname path
.
```

I can try looking into a suitable way to do this tomorrow…

sampsyo:
I gave this a try:

```
#!/bin/sh
script=`readlink $0 || echo $0`
here=`dirname $script`
echo $here
```

And it does avoid the error. It works as long as the symbolic link is absolute, but if it's relative, `readlink` will still give you a path that's relative to the symbolic link's location—so `$here` is a lie from the perspective of the working directory. I don't see an obvious way around that…

rachitnigam:
Hm, that seems like a reasonable restriction. I've updated the script to use your snippet. Merge if you think it'll work fine.

sampsyo:
Seems fine, at least for now! A solution that is vaguely frustrating but would work could use Python's `realpath` function:

```sh
python3 -c 'import os, sys; print(os.path.realpath(sys.argv[1]))' /path/to/file/that/might/be/a/link
```

PR #112, opened 2019-03-04
Add Vim stuff to docs

tedbauer:


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
rachitnigam: Why was this committed?

tedbauer: Removed!

rachitnigam: The `website/yarn.lock` is important. Just remove the empty one you created.

PR #113, opened 2019-03-05
Clean-up examples

sa2257:
Update examples to only carry benchmarks that synthesize on Buildbot (Seashell `gemm` does run into timing issues).


examples/machsuite/mach-common/Makefile
```diff
@@ -1,7 +1,9 @@
 
 BENCHMARKS=\
-	../mach-gemm_ncubed \
-	../mach-stencil2d
+	../mach-gemm_ncubed 
+#	../mach-stencil2d
```
rachitnigam: Are these meant to be commented out? If so, can you remove them in this pull request and add them as you add benchmarks?

rachitnigam:
@sa2257 Regarding our previous conversation about moving benchmarks to a new repo. Should we just move them there now? Is there a good argument for keeping them in this repo?

sampsyo:
Thanks for cleaning this up!

Is there anything MachSuite specific about the `mach-common` directory? It looks like that just contains a Makefile that submits stuff to Buildbot for now. That seems (a) like a task that's generically about benchmarking, not about MachSuite in particular, and (b) like the sort of thing that can live in our new, private repository for benchmarking _infrastructure_. So unless there are other plans for it, let's just delete that directory.

rachitnigam:
Moved examples/ to https://github.com/cucapra/fuse-benchmarks

PR #114, opened 2019-03-05
External functions

rachitnigam:
Add support for specifying external functions:
```
def extern print_vector(a: float[10]);
print_vector(a)
```

`extern` functions are only checked using their function signatures. Different backends can decide to do different things for each `extern`. `vivado` and `c++` remove all `extern` defs but keep calls to them. This allows us to use functions provided by header files.

Other backends might do different things for each `extern`. For example, a backend might decide to provide the implementation of `print_vector` in its implementation.

### DIFF
--- a/docs/binders.md
+++ b/docs/binders.md
@@ -73,7 +73,7 @@ declared using the `extern` keyword.
 def extern print_vector(c: float[4]);
 ```
 
-Backends can either provide their implementations for `externs` (for example
+Backends can either provide their implementations for `extern`s (for example
 to implement printing when running on CPUs) or erase the calls completely.
 
 For more details on the exact hardware semantics see here **TODO**.
--- a/docs/binders.md
+++ b/docs/binders.md
@@ -66,4 +66,14 @@ def foo(x: bit<10>, y: bit<10>) {
 }
 ```
 
+External functions, such as functions provided by header files, can be
+declared using the `extern` keyword.
+
+```
+def extern print_vector(c: float[4]);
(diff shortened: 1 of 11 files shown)


docs/binders.md
```diff
@@ -66,4 +66,14 @@ def foo(x: bit<10>, y: bit<10>) {
...
 ```
 
+External functions, such as functions provided by header files, can be
+declared using the `extern` keyword.
+
+```
+def extern print_vector(c: float[4]);
+```
+
+Backends can either provide their implementations for `externs` (for example
```
sampsyo: This is ridiculously pedantic, but:

    `extern`s

seems more correct than:

    `externs`

src/main/scala/Syntax.scala
```diff
@@ -173,7 +173,7 @@ object Syntax {
   case object CEmpty extends Command
 
   sealed trait Definition extends Positional
-  case class FuncDef(id: Id, args: List[Decl], body: Command) extends Definition
+  case class FuncDef(id: Id, args: List[Decl], bodyOpt: Option[Command]) extends Definition
```
sampsyo: A comment here about why the body is optional seems useful.

docs/binders.md
rachitnigam: fixed.

src/main/scala/Syntax.scala
rachitnigam: Fixed.

PR #115, opened 2019-03-07
Import statements and run testing harness

rachitnigam:
Two changes:
- Support for `import` statements:
```
import "printer.h" {
  def extern print_vect(a: float[4])
}
```
which gets compiled to:
```
#include "printer.cpp"
```
and always extern calls to `print_vect` in the program.

- `-i` flag for `run` mode which specifies location to headers. For example:
```
fuse run matadd.fuse -o matadd.fuse -i /usr/bin -i /usr/local/bin
```
gets turned into the call
```
$(CXX) -I /usr/bin -i /usr/local/bin
```

This separation will allow us to write `run` tests (Upcoming on a new branch).


src/main/scala/Main.scala
```diff
@@ -43,7 +43,12 @@ object Main {
         opt[String]('o', "outfile")
           .required()
           .action((f, c) => c.copy(output = Some(f)))
-          .text("name of the output artifact."))
+          .text("Name of the output artifact."),
+        opt[String]('i', "include")
+          .optional()
+          .unbounded()
+          .action((i, c) => c.copy(includes = i :: c.includes))
+          .text("Include location for header for CXX compilation."))
```
sampsyo: Seems fine, but if you'll permit a slippery slope argument: it wouldn't do to duplicate all the `gcc` CLI flags as `fuse` flags. A lighter-weight but less convenient option would have a flag that means "pass this to the underlying compiler." Like `-x "-I path/to/header"`.

src/test/scala/TypeCheckerSpec.scala
```diff
@@ -1003,4 +1003,16 @@ class TypeCheckerSpec extends FunSpec {
       }
     }
   }
+
+  describe("Imports") {
+    it("adds externs into type checking scope") {
+      typeCheck("""
+        import "print.cpp" {
```
sampsyo: I know these are just the tests, but `#include`ing *.cpp files is rarely what you want—maybe use `print.h` instead.

rachitnigam: Fixed.

src/main/scala/Main.scala
rachitnigam: Yeah, `-i` feels hacky. `-x` makes more sense. There is no good reason to choose one over the other.

rachitnigam: Fixed.

rachitnigam:
I've added the commits for the run testing harness in this too. Take a look at `docs/testing.md` to see how `RunTest` work.

docs/testing.md
```diff
@@ -0,0 +1,71 @@
...
+}
+
+decl gold: ...
+...
+
+assert(res == gold)
+```
+
+Next, create a file with the name `<src>.data.json` and add a json file with
+the arguments to the kernel as specified in ["Generating executables"](cpp-runnable.md).
```
sampsyo: Maybe worth mentioning that you can include `gold` (i.e., the expected output) as "just another argument"?

src/main/scala/Main.scala
```diff
@@ -43,7 +43,12 @@ object Main {
         opt[String]('o', "outfile")
           .required()
           .action((f, c) => c.copy(output = Some(f)))
-          .text("name of the output artifact."))
+          .text("Name of the output artifact."),
+        opt[String]('x', "--compiler-opt")
+          .optional()
+          .unbounded()
+          .action((x, c) => c.copy(compilerOpts = x :: c.compilerOpts))
+          .text("Include location for header for CXX compilation. Can be repeated."))
```
sampsyo: Description string is out of date.

docs/testing.md
```diff
@@ -0,0 +1,71 @@
...
+The file test harness simply pulls in tests under `src/test/{should-compile,should-fail}`.
+As the names suggest, the first directory contains tests that should compile with
+the type checker while the second contains programs that should fail.
+
+Simply add a Fuse program under the directory to get them to run.
+
+## Run tests
+
+This harness is used to check the functional correctness of the emitted code.
+It creates a temporary directory `_runTestTmp` to store intermediary C++ files.
```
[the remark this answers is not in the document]
rachitnigam: The only benefit is that when tests fail, you can look at the C++ generated under that deterministically generated directory. Thoughts?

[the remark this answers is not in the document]
sampsyo: Got it. Both options seem reasonable, (cd'ing to /tmp would be frustrating), but since the goal is convenience, I recommend calling it something simple like `_test` or `_build` or `_tmp`.

sampsyo:
```
alias ragit='git commit -a -m x ; git push'
```

More seriously: good luck, and if you can figure out how I can give you access to the Circle VM directly, please let me know!

rachitnigam:
It's fixed! Also, I have SSH access to it.

docs/testing.md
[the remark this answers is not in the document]
rachitnigam: Renamed to `_test`.

docs/testing.md
rachitnigam: Fixed.

src/main/scala/Main.scala
rachitnigam: Fixed.

PR #119, opened 2019-03-10
Buildbot UX fixes

rachitnigam:
Three changes:

- Extract state information out.
- Rename some titles and change format for date.
- Add color labels for states

state-labels

### DIFF
--- a/buildbot/buildbot/server.py
+++ b/buildbot/buildbot/server.py
@@ -46,9 +46,9 @@ def _datetime_filter(value, withtime=True):
         return ''
     dt = datetime.fromtimestamp(value)
 
-    fmt = '%B %d, %Y'
+    fmt = '%m/%d/%y'
     if withtime:
-        fmt += ', %I:%M %p'
+        fmt += ', %H:%M'
     return _unpad(dt.strftime(fmt))
 
 
--- a/buildbot/buildbot/server.py
+++ b/buildbot/buildbot/server.py
@@ -7,6 +7,7 @@ from . import workproc
 from .db import JobDB, ARCHIVE_NAME, CODE_DIR, NotFoundError
 from datetime import datetime
 import re
+from . import state
 
 # Our Flask application.
 app = flask.Flask(__name__, instance_relative_config=True)
@@ -103,7 +104,7 @@ def add_job():
             return 'invalid extension {}'.format(ext), 400
 
         # Create the job and save the archive file.
-        with db.create('uploaded', config) as name:
+        with db.create(state.UPLOAD, config) as name:
             file.save(ARCHIVE_NAME + ext)
         notify_workers(name)
 
@@ -111,7 +112,7 @@ def add_job():
         code = request.values['code']
 
         # Create a job and save the code to a file.
-        with db.create('unpacked', config) as name:
+        with db.create(state.UNPACK_FINISH, config) as name:
             os.mkdir(CODE_DIR)
             with open(os.path.join(CODE_DIR, 'main.ss'), 'w') as f:
                 f.write(code)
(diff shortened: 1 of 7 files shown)


buildbot/buildbot/state.py
```diff
@@ -0,0 +1,12 @@
...
+UPLOAD = "Uploaded",
+UNPACK = "Unpacking",
+UNPACK_FINISH = "Unpacked",
+COMPILE = "Compiling",
+COMPILE_FINISH = "Compiled",
+HLS = "HLS",
+HLS_FINISH = "HLSed",
+RUN = "Running",
+DONE = "Done"
+FAIL = "Failed"
```
sampsyo: Let's keep these strings as all lower-case, please. (They're meant as internal status strings, not part of a polished UI. If we want to make the UI version of these strings better, let's make those separate. Keeping the same names as before will also make currently-terminated job data we have lying around still look right. Also, capitalized CSS class names look weird. :smiley:)

Also, these lines do not need trailing commas.

buildbot/buildbot/templates/base.html
```diff
@@ -11,7 +11,8 @@
         {% block head %}{% endblock head %}
     </head>
     <body>
-        <h1>Seashell Buildbot</title>
+        <h1>Buildbot</title>
+        <hr/>
```
sampsyo: This hr looks a tiny bit clunky to me… I think I'd rather just put a border-bottom on the h1. I can fiddle with that though.

buildbot/buildbot/server.py
```diff
@@ -46,9 +47,9 @@ def _datetime_filter(value, withtime=True):
         return ''
     dt = datetime.fromtimestamp(value)
 
-    fmt = '%B %d, %Y'
+    fmt = '%m/%d/%y'
     if withtime:
-        fmt += ', %I:%M %p'
+        fmt += ', %H:%M'
```
sampsyo: This is definitely more legible than the old one, but if we're going for filename-like mental sortability, how about something like `%Y-%M-%d %H:%M`, i.e., approximately [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)?

sampsyo:
Belatedly, it occurs to me that the colors might be most effective if they just represented the "high-order bit" of the status: in progress, failed, or successful (done). Yellow, red, and green could capture those.

buildbot/buildbot/state.py
rachitnigam: Is there a way to easily import a map from internal string to UI strings in Flask templates?

sampsyo: A Python dictionary? :smiley: You can do `{{ state_strings[state] }}`, for example.

rachitnigam: Can I also write `import state` in the template? I understand how to use python dicts, but wanted to keep all state data in state.py.

sampsyo: There are two good ways to do this: pass along `state_strings` as an additional parameter to the template (from `server.py`), or set up the templating engine so that `state_strings` is in the [global namespace](http://jinja.pocoo.org/docs/2.10/api/#the-global-namespace).

sampsyo: But, the human-readable names of states might best belong in `server.py` because it's bowser-UI-related…

buildbot/buildbot/templates/base.html
rachitnigam: Fixed.

buildbot/buildbot/server.py
rachitnigam: Fixed.

buildbot/buildbot/state.py
rachitnigam: Fixed.

rachitnigam:
Fixed the comments. Regarding the last comment, can we try out the colorful labels for a while and revert to the three color format if it's too noisy?

sampsyo:
Of course; sounds good.

PR #120, opened 2019-03-10
Use name of archive to generate buildbot job name

rachitnigam:


sampsyo:
Is the goal here to make it more obvious which job is which when looking at the listing on the main page? If so, a nicer way to accomplish this with fewer drawbacks would be to put the original filename into the job's metadata (i.e., configuration). Then the homepage could have another column (after the ID column) that shows this filename (or nothing if the field does not exist).

rachitnigam:
There are two goals. The first is imwhat you stated. The second one is getting back names of benchmarks by the extraction script. Would it be easy to return the name of the benchmarks in extract.py using this method?

sampsyo:
Certainly! If you request `/jobs/<name>`, you get a JSON dictionary with all of that data, so you could pull the string right out of there.

rachitnigam:
It seems like `hw_basename` is already extracted for the status page. Should I just use that?

sampsyo:
Yep, that seems useful enough.

PR #122, opened 2019-03-12
Fixing CSeq type checking rule

### LINKED ISSUE #51: Type checking rule for `---`
We have the following type checking rule for `e1 --- e2`.

```
G, e1 --> G', t1    G, e2 --> G'', t2  Gf = G' + G''
---------------------------------------------------------------    where G' + G'' = G''
                G, e1 --- e2 --> Gf, t2
```

The definition of `+` on `G` is incorrect for this case:
```
{
  let x = 1;
  ---
  let y = 2;
};
x + y // error with `x` is unbound.
```

Figure out the right `+` and implement it.

apurvakoti:
Fixes #51 
Creates a new scope in the environment; typechecks `c1`; extracts any new bindings from this typechecking; adds the new bindings to the initial environment and typechecks `c2`.

### DIFF
--- a/src/main/scala/TypeCheck.scala
+++ b/src/main/scala/TypeCheck.scala
@@ -360,9 +360,10 @@ object TypeChecker {
     case CExpr(e) => checkE(e)._2
     case CEmpty => env
     case CSeq(c1, c2) => {
-      val _ = checkC(c1)
-      val e2 = checkC(c2)(env, rres)
-      // FIXME(rachit): This should probably be intersection of e1 and e2
+      val newscope = env.addScope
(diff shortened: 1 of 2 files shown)


rachitnigam:
The implementation looks correct to me! Can you add a few more tests that nest the `---` under unrolled loops.

Also, two minor nits:
- The tests belong under "sequential composition"
- The strings for `it` should start with small letters.

rachitnigam:
@apurvakoti can you add the fixes so we can merge?

rachitnigam:
LGTM.

PR #131, opened 2019-03-14
Integration data extraction

sa2257:
Mainly to add data extraction scripts, data_extract.py and minor changes to worker.py

Also has merged with buildbot-hls-simplify and master.


rachitnigam:
This branch history seems to be in some weird state where most of the changes have already been merged it. We'll have to clean it up before merging.

Three high level points:
- We should separate out runtime information capturing into a different PR (I know this remote doesn't have those changes yet, but just in case).
- Use the libraries I rewrote in fuse benchmarks for parsing in data. They are more robust at parsing RPT data.
- Let's hold off on merging the buildbot side result parsing till we have more benchmarks and are sure that the aforementioned libraries are in some reasonable stable state. Iterating server side is easier for now.

sa2257:
Buildbot going on it's own way in #210 extraction is probably better done locally, even though having data extracted to logs can be useful for a job snapshot and debugging.

PR #134, opened 2019-03-16
Define withScope and move ReqResources into typeEnv

rachitnigam:


src/main/scala/TypeEnv.scala
```diff
@@ -54,18 +51,26 @@ object TypeEnv {
...
     /**
      *  Remove the topmost scope from the environment. The bindings and capabilities
-     *  bound in this scope are removed from the context.
+     *  bound in this scope are removed from the context. This method is
+     *  only used for interface implementation and should not be used directly.
      *  @returns A new environment without the topmost scope, A Scope containing
      *           all bindings in the topmost scope, A scope containing capability mappings.
      */
-    def endScope: (Environment, Scope[Id, Info], Scope[Expr, Capability])
+    def endScope: (Environment, Scope[Id, Info])
```
rachitnigam: @sampsyo Abstraction question: I don't want this function to ever be used but I need to expose it in the `Environment` interface because the implementation of `withScope` returns the `Environment` trait. 

Is there a clean way of doing this (either here or in OCaml)? Also, does this "pattern" have a name?

sampsyo: To summarize, it sounds like the central issue is that the primary interface you want to expose to environments is `withScope`, which runs whatever `Environment => Environment` function you like in a fresh scope. In ordinary operation, no "clients" should be manufacturing their own environments (by initializing new `Environment` value from scratch) or calling `endScope`; the only way they should be creating/manipulating environments should be `withScope`.

So in an ideal world, those functions that get passed into `withScope` should only have access to some "limited" version of the interface that only exposes the data members and a `withScope` function for recursion. Notionally, say we had some trait called `E` that just exposed these operations. The function argument to `withScope` would then have type `E => E`. The trouble, then, would be that `withScope` would get an `E` back from its `inScope` call, and it couldn't be certain that this was actually an `Environment`, i.e., something it could call `endScope` on. A squirrelly function could have initialized its own `E` from scratch with some other implementation!

I think the natural way around this would be to make those `Environment => Environment` functions polymorphic. That is, they'd have to be `T => T` for some `T` that extends `E`. The `withScope` function would parameterize this function with `Environment`, thereby getting a guarantee that the returned value is actually an `Environment` while never exposing this fact to the `T => T` function.

Does that make any sense?

rachitnigam: Hm, I don't see how that would work. If we have:
```
withScope[T :> E](func: T => T)
```
A random function can still access methods like `endScope` because it knows that type `T` extends `E` and `E` provides the required methods.

It's unclear that what I'm asking is actually doable. Environments need to implement a way to end their scope somehow. The useful part of `withScope` is the ability to track the resources in the current scope without adding an explicit stack in the environment. If we didn't have `withScope`, to implement the resource tracking, the call to an `endScope` would need to know how many resources were added by the last `addScope`. `withScope` doesn't have this issue because it can simply pop the same number of resources as its parameter.

sampsyo: Yeah, I'm also not 100% sure it's possible, but to push it one step further: `E` would need to be a *different* trait that only exposes `withScope` (and the data accessors). Then the proper environment implementation, which includes `endScope`, would need to extend that trait but remain hidden from clients.

rachitnigam: I extracted out the scope management functionality into `ScopeManager`, but seem to be running into the same problem where the implementation of `endScope` leaks through. There might be some convoluted type class + inheritance implementation that does what I want, but for now, I'm going to leave `addScope` and `endScope` in.

PR #141, opened 2019-03-18
Bounds checking

sgpthomas:


src/main/scala/BoundsCheck.scala
```diff
@@ -0,0 +1,45 @@
...
+object BoundsChecker {
+
+  def check(p: Prog) : Unit = {
+    checkC(p.cmd)
+  }
+
+  private def checkE(e: Expr) : Unit = {
+    e match {
+      case e@EArrAccess(id, idxs) => e.dims match {
+        case Some(dims) => (idxs.map(_.typ) zip dims).foldLeft(())((_, ele) => ele match {
```
rachitnigam: Scala style: Use `map` over options instead of case matching.

rachitnigam: Another style thing. You can rewrite:
```scala
(foo => foo match {
  case (x, y) => ...)
```
to:
```scala
({ case (x, y) => ... })
```

src/main/scala/Syntax.scala
```diff
@@ -112,12 +112,15 @@ object Syntax {
...
     }
+    var typ: Option[Type] = None
   }
   case class EInt(v: Int, base: Int = 10) extends Expr
   case class EFloat(f: Float) extends Expr
   case class EBool(v: Boolean) extends Expr
   case class EBinop(op: BOp, e1: Expr, e2: Expr) extends Expr
-  case class EArrAccess(id: Id, idxs: List[Expr]) extends Expr
+  case class EArrAccess(id: Id, idxs: List[Expr]) extends Expr {
+    var dims: Option[List[(Int, Int)]] = None
```
rachitnigam: I don't think you need this. Each `Id` in the program should be decorated with it's type. The `id` here should get a `TArray` type with all the relevant information you want.

src/main/scala/Syntax.scala
```diff
@@ -112,12 +112,15 @@ object Syntax {
       case _:EVar | _:EArrAccess => true
       case _ => false
     }
+    var typ: Option[Type] = None
```
rachitnigam: What is this used for?

sgpthomas: I need to know the type of each index while bound checking. The type checking for this happens in `TypeChecker.consumeBanks` and so I save the type of each index expression.

src/main/scala/Syntax.scala
sgpthomas: You're right :+1:

src/main/scala/Syntax.scala
sampsyo: As usual, this opens up a bigger discussion. The general idea here is *type elaboration*, which is something most compilers need to do—i.e., associate a type with every AST node (or at least, with every expression and every declaration). It's necessary because most downstream passes need to know the types of everything. Code generation typically needs to know about types, and indeed that's why identifiers already have a type field (as Rachit mentioned some time ago: https://github.com/cucapra/seashell/pull/46/files#r254274169). And clearly bounds checking needs to know the types also.

So in this or some subsequent PR, we should address that problem head-on. That is, we should invent some data structure for holding the elaborated types, and both codegen and bounds checking should use that. It should replace the one-off type field for idents.

That raises the question about the right way to store the elaborated types. [This blog post by Edward Yang](http://blog.ezyang.com/2013/05/the-ast-typing-problem/) gives a succinct overview of the typical options, which are all imperfect in their own way. The `typ: Option[Type]` strategy is called "Nullable field" in the blog post.

sgpthomas: I think this would be better in its own PR. Another related topic is a nice way to write new passes over the tree. For example, for the bounds checking, I had to write a lot of uninteresting code as boilerplate:
```
  private def checkC(c: Command) : Unit = c match {
    case CPar(c1, c2) => checkC(c1) ; checkC(c2)
    case CSeq(c1, c2) => checkC(c1) ; checkC(c2)
    case CLet(_, _, exp) => checkE(exp)
    case CView(_, _) => ()
    case CIf(cond, tbranch, fbranch) => checkE(cond) ; checkC(tbranch) ; checkC(fbranch)
    case CFor(_, par, combine) => checkC(par) ; checkC(combine)
    case CWhile(cond, body) => checkE(cond) ; checkC(body)
    case CUpdate(lhs, rhs) => checkE(lhs) ; checkE(rhs)
    case CReduce(_, l, r) => checkE(l) ; checkE(r)
    case CExpr(e) => checkE(e)
    case CEmpty => ()
  }
```
It would be nice if I could simply target the branches where I want to do something.
 This depends a little bit on how many passes we anticipate writing.

rachitnigam: Yeah, that's definitely a common issue when writing compiler. In the OCaml codebase, I implemented [a visitor class](https://github.com/cucapra/seashell/blob/3d25f46e67f2512aba707bdf9daa9dc5d7c5f3aa/src/ast_visitor.ml) for the AST. That allowed us to write some [clean passes](https://github.com/cucapra/seashell/blob/3d25f46e67f2512aba707bdf9daa9dc5d7c5f3aa/src/flatten_seq.ml).

The code under RewriteViews.scala has beginning of a monadic visitor. It would be nice to extract the logic into something reusable.

src/test/should-compile/index-bounds.fuse
```diff
@@ -0,0 +1,6 @@
+
```
rachitnigam: By default, the small example should go in TypeCheckerSpec.scala which allows us to track errors thrown and values expected in a more granular way. `should-compile` and `should-fail` are mostly for larger programs or benchmarks.

sgpthomas:
Addressed the comments

src/main/resources/headers/parser.cpp
```diff
@@ -78,6 +78,11 @@ vector<vector<T>> to_order_2_tensor(picojson::array &arr, unsigned int row_dim,
   return ret;
 }
 
+template <typename T>
+inline vector<T> flatten_matrix(vector<T> vect) {
+  return vect;
+}
+
```
rachitnigam: This has been fixed on master.

rachitnigam:
This looks really good! Let's merge it in person after our Thursday meeting. In the meantime you can look into:
- Rebasing on master and resolving any conflicts.
- Using [scala-logging](https://github.com/lightbend/scala-logging) to emit compiler warnings instead of `println`. We can even have nice color labels using the trick [used in Compiler.scala](https://github.com/cucapra/seashell/blob/600e9fa128dddb60f9f2254449f24fa2c11b5e7a/src/main/scala/Compiler.scala#L21).
  - If you do end up adding it, we can also have a `-d`, `--debug-level` parameter passed to the CLI that controls the warnings emitted by the compiler.

sgpthomas:
I looked at scala-logging and it seems like its just a simple wrapper over java logging. I found a library called scribe that is native to scala and gives you nicer control. Here's a [blog](https://medium.com/airframe/airframe-log-a-modern-logging-library-for-scala-56fbc2f950bc) that discusses this in more detail.

For warnings, it would be nice to point to synax locations so I will look into seeing if this is possible.

rachitnigam:
Oh neat! Some considerations:
- I eventually want to be able to compile to JavaScript using Scala.js. A native library might make it slightly harder. Scala-logging claims that you can switch out the underlying library "easily".
- Syntax locations are already captured in every Syntax object that extends `Positional`. I think the array access you want to show the location of can already be done. Take a look at `Errors.scala` to see how errors report locations.

src/main/scala/TypeCheck.scala
```diff
@@ -74,26 +74,29 @@ object TypeChecker {
...
           if (dims(i)._2 != 1) throw InvalidDynamicIndex(id, dims(i)._2)
           else env2.update(id, env(id).consumeBank(i, 0)) -> bres * 1
-
         case (t, _) => throw InvalidIndex(id, t)
       }
     })
 
   private def checkLVal(e: Expr)(implicit env: Environment, rreq: ReqResources) = e match {
-    case EArrAccess(id, idxs) => env(id).typ match {
+    case e@EArrAccess(id, idxs) => env(id).typ match {
```
rachitnigam: This change doesn't seem necessary anymore.

src/main/scala/TypeCheck.scala
```diff
@@ -74,26 +74,29 @@ object TypeChecker {
...
       // This only triggers for r-values. l-values are checked in checkLVal
       case TArray(typ, dims) => {
         if (dims.length != idxs.length) {
           throw IncorrectAccessDims(id, dims.length, idxs.length)
         }
+        // println(idxs.map(ex => ex.typ) mkString ", ");
```
rachitnigam: Comment is not needed.

src/test/scala/TestUtils.scala
```diff
@@ -11,5 +11,10 @@ object TestUtils {
 
   def parseAst(s: String) = FuseParser.parse(s)
   def typeCheck(s: String) = TypeChecker.typeCheck(FuseParser.parse(s))
+  def boundsCheck(s: String) = {
```
rachitnigam: Just redefine typeCheck to have the bounds check. No need for a new method.

PR #144, opened 2019-03-19
Remove -M options for sds++ tasks

rachitnigam:


sampsyo:
I checked out [the docs](https://clang.llvm.org/docs/ClangCommandLineReference.html), and all of these are `-M`-related options for generating Makefile dependencies:

- `-MMD`: Generate user header dependencies for writing to a separate file.
- `-MP`: Include Make `.PHONY` targets in the dependency file for the targets.
- `-MF"vsadd.d"`: Write the dependency stuff to this file.

So yep! I can pretty confidently say we don't need these in the compiler flags.

Issue #137, opened 2019-03-16
No system for emitting compiler warnings


sgpthomas:
It seems like there should be a standardized way to emit compiler warnings along with command line options to enable/disable them

rachitnigam:
A generic logging framework like [scala-logging](https://github.com/lightbend/scala-logging) is usually the right way to do this.

That way, we can implement debugging information as just another “logging level” in the compiler. The CLI can have a simple argument to change the logging level.

PR #145, opened 2019-03-19
Added scribe as a logging system, addresses #137

sgpthomas:
This seems like a nice simple logging framework. It advertises seamless support with Scala.js. It is pretty simply to add logging for new types of objects. I've done this for Positionals so that you can log/warn fuse source locations.

### DIFF
--- a/src/main/scala/Main.scala
+++ b/src/main/scala/Main.scala
@@ -2,6 +2,8 @@ package fuselang
 
 import java.nio.file.{Files, Path}
 import java.io.File
+import scribe.{Logger, Level}
+import scribe.format._
 
 import Utils._
 
@@ -36,6 +38,10 @@ object Main {
       .action((b, c) => c.copy(backend = toBackend(b)))
       .text("Name of the backend to use. Default backed is vivado.")
 
+    opt[Unit]('d', "debug")
+      .action((_, c) => c.copy(logLevel = Level.Debug))
+      .text("Enables debug output")
+
     cmd("run")
       .action((_, c) => c.copy(mode = Run, backend = toBackend("c++")))
       .text("Generate a runnable object file. Assumes GCC and required headers are available. Implies mode=c++.")
@@ -53,10 +59,15 @@ object Main {
 
   def main(args: Array[String]): Unit = {
 
+
     type ErrString = String
 
     parser.parse(args, Config(null)) match {
       case Some(conf) => {
+        val myFormatter: Formatter = formatter"[$levelColored] $message$newLine"
+        Logger.root.clearHandlers().withHandler(formatter = myFormatter,
+          minimumLevel = Some(conf.logLevel)).replace()
+
         val path = conf.srcFile.toPath
 
         val prog = Files.exists(path) match {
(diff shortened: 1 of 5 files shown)


src/main/scala/Main.scala
```diff
@@ -53,10 +59,15 @@ object Main {
 
   def main(args: Array[String]): Unit = {
 
+
     type ErrString = String
 
     parser.parse(args, Config(null)) match {
       case Some(conf) => {
+        val myFormatter: Formatter = formatter"[$levelColored] $message$newLine"
```
rachitnigam: Move the logger setup code to either (1) start of main, or (2) a separate function in Main.scala.

src/main/scala/Syntax.scala
```diff
@@ -1,11 +1,19 @@
...
 object Syntax {
 
   import Errors._
 
+  /** Makes all positionals logable by scribe */
+  implicit object PositionalLoggable extends Loggable[(String,Positional)] {
+    override def apply(value: (String, Positional)) = {
+      new output.TextOutput(s"${value._1}\n${value._2.pos.longString}")
+    }
+  }
```
rachitnigam: How is this used? Do I simply do `logger.warn(expr)`? Can you write an example (in the github thread) of how you use it?

sgpthomas:
It's simply `scribe.warn("warning")`. Similarly for, `debug`, `info`, `error`. You also don't have to print strings, so for example I added code to print `(String, Positional)`, and so `scribe.warn(("some message", expr))` will work.

rachitnigam:
Is it possible to simply use the `withPos` function defined for Errors instead of the implicit? In general, I shy away from having global implicit (which this seems to be) and implicit classes on common things like pairs.

sgpthomas:
Yeah it's definitely possible to do without the implicit class. You can just use something like
`scribe.warn(Errors.withPos("hi there", Some(expr.pos)));`. Maybe I'm misunderstanding something but this implicit class isn't defining a new method on tuples, it's extending the `Loggable` trait defined in `scribe`. Basically, it looks like the `scribe.info` and related functions take in a `Loggable[M]` value. The `Loggable` trait then has information  about how to transform `M` into something loggable. So in this case the implicit class is providing a new way to print out `(String, Positional)` values. I'm not seeing why this is so bad, though I understand why it's bad in general to add a lot of implicit functionality.

sgpthomas:
This is the library code for making strings loggable
```scala
  implicit object StringLoggable extends Loggable[String] {
    def apply(value: String): LogOutput = new TextOutput(value)
  }
```

rachitnigam:
Oh, I see. Seems like it's a [Scala typeclass](https://blog.scalac.io/2017/04/19/typeclasses-in-scala.html). The methods here:

https://github.com/outr/scribe/blob/544886b43624acd7801461abbad2701e1828d644/core/shared/src/main/scala/scribe/LoggerSupport.scala#L17

try to view the parameters with a specific typeclass. I'm fine with having that logic in, however, I don't think it belongs in the Syntax.scala file. Make a Logger.scala for the logging configuration and the implicit objects and we can use it to store logging specific stuff.

rachitnigam:
LGTM!

Issue #32, opened 2018-10-03
Buildbot tracker


sampsyo:
TODO:
---

For benchmark harness (**High Priority**)
- [ ] Buildbot should extract data using the data extraction scripts after the synthesis step.
- [x] Run task should upload required data files to Areesh.
- [ ] First order support for F1 instances.
---
* [x] Logs should be kept in a proper text file, `log.txt`, instead of in the JSON job info dictionary. As a consequence, running commands could "stream" to the log, letting us see in-progress output before the command is done executing.
* [ ] Job page should extract all log lines that have the word "warning". These usually seem to indicate when a particular pragma is being ignored.

DONE:
---

* [x] The job page should not have the full file listing and log. (These get way too big in normal runs.) Link to separate pages to view these.
- [x] Buildbot should detect if there is a Makefile and run `make`.
* [x] Finish fixing the Dockerfile to work with opam2 so it can actually run on Gorgonzola again. (@sampsyo is working on this.)
* [x] Merge the workproc branch described in https://github.com/cucapra/seashell-gorgonzola/issues/1 to fix reliability issues.
* [x] Make the buildbot actually run things on the Zynq board. (@sa2257 is working on this.)
* [x] Add a no-Seashell mode to just HLSify and execute plain C programs.
* [x] Add an option to just run the Vivado hardware resource estimation instead of doing the full synthesis flow.

* [x] Add the ability to change the status of a job. This is really useful for debugging, when you want to take an old job and "rewind" it to an earlier stage that failed before. (It's currently possible to do this by hand by editing the JSON job data file, but then you have to give the server a kick somehow to pick up the new status.)

- [x] Generate job name based on the `cpp` or `fuse` file in the provided archive.
- [x] Change the format for "started" to "%MM/%DD, %TIME". Also the time currently given is 3 hours late.
- [x] Add color labels for job states to quickly figure the status of recent jobs.

sampsyo:
And here's a pie-in-the-sky idea:

* [ ] Get the buildbot running on an [Amazon F1](https://aws.amazon.com/ec2/instance-types/f1/) instance, which includes [Xilinx SDSoC for free](https://github.com/aws/aws-fpga/blob/master/SDAccel/README.md). This would get us access to a somewhat bigger FPGA and make the setup accessible to other people.

sa2257:
- [ ] Fix Areesh file copy bug
- [ ] Submit jobs with multiple hardware functions
[...]

PR #147, opened 2019-03-20
buildbot: Move jobs' file lists and logs to separate subpages

sampsyo:
As mentioned in #32, this makes the main job pages shorter and therefore less annoying to use. They link to separate pages for the file list and the log. The main job page still shows the last few messages from the log, but not the whole thing.

Here's what the new job page looks like:

Screenshot 2019-03-19 21 05 30

### DIFF
--- a/buildbot/buildbot/server.py
+++ b/buildbot/buildbot/server.py
@@ -204,6 +204,16 @@ def show_job(name):
     )
 
 
+@app.route('/jobs/<name>/log.html')
+def job_log(name):
+    job = _get(name)
+
+    return flask.render_template(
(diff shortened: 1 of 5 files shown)


rachitnigam:
Seems like this shows the last 4 messages which is great!

Do you think implementing the log as a txt file instead of a json dict is easy enough to do in this PR? If you point me to a location, I can implement it.

sampsyo:
Great!

I don't think that will be quite so easy, unfortunately—it will involve setting up the command runner to pipe output to a file, and making sure we still have a way to record when commands start. It will take a little more attention.

PR #149, opened 2019-03-21
Parsing for records

### LINKED ISSUE #148: Add support for parsing records in C++ runnable backend
In my `seashell/` dir, I ran `./fuse run bfs.fuse -o bfs.cpp`.

Contents of `bfs.fuse`:
```
record edge_t {
  dst: bit<32>
}

record node_t {
  edge_begin: bit<32>;
  edge_end: bit<32>
}

decl nodes: node_t[50];
decl edges: edge_t[50];
decl starting_node: bit<32>;
decl level: bit<32>[50];
decl level_counts: bit<32>[50];

let cnt = 0;

// Temporary variables for serial updates
let tmp_begin:bit<32> = 0;
let tmp_end:bit<32>   = 0;
let tmp_dst:bit<32>   = 0;
let tmp_level:bit<32> = 0;

let MAX_LEVEL = 10;

let tmp_l:bit<32> = 0;

for (let horizon = 0..50) {
  cnt := 0;
  for (let n = 0..50) {
    tmp_l := level[n];
    ---
    if (tmp_l == horizon) {
      tmp_begin := nodes[n].edge_begin;
      ---
      tmp_end := nodes[n].edge_end;
      ---
      for (let e = 0..50) { // FIXME: find upper bound
        if (e >= tmp_begin && e <= tmp_end) {
          tmp_dst := edges[e].dst;
          tmp_level := level[tmp_dst];
          ---
          if (tmp_level == MAX_LEVEL) {
            level[tmp_dst] := horizon + 1;
	          ---
	          cnt := cnt + 1;
          }
        }
      }
    }
  }
}
```
The compiler emitted this error:
`[Impossible] Cannot parse type node_t[50 bank 1] with CppRunnable backend. This should never trigger. Please report this as a bug.`

rachitnigam:
Fixes #148.

Rewrite the parsing library to use [json.hpp](https://github.com/nlohmann/json) which provides cleaner interfaces.
- Each record def generate parsing helpers which are used by the json library to automagically convert json elements into records.
- Cleaned up some old code in CppRunnable.scala.


rachitnigam:
Ready for review.

docs/cpp-runnable.md
```diff
@@ -4,7 +4,18 @@ id: cpp-runnable
...
-can be used to test Fuse code without going throught a synthesis workflow.
+can be used to test Fuse code without going through a synthesis workflow.
+
+## Limitation
+
+The backend is meant to be used for functional testing and as such has a few
+important limitations.
+
+1. Exact bitwidth integer computation is not supported. Sized integers get
+   compiled to C++ `int`.
```
sampsyo: Could be worth making a separate issue about addressing this limitation, which shouldn't be too hard (using [ac_int](https://www.mentor.com/hls-lp/downloads/ac-datatypes), perhaps).

docs/cpp-runnable.md
```diff
@@ -4,7 +4,18 @@ id: cpp-runnable
...
+2. `unroll`s do not create any form of parallelism. The annotation is compiled
+   away after typechecking.
+3. Functions are not inlined.
```
sampsyo: These two limitations, unlike the first one, should not affect semantics (only performance).

rachitnigam: Added clarification.

docs/cpp-runnable.md
rachitnigam: Adding this to the low priority tracker. This is slightly non trivial because the Json library can't extract the type directly using `get_arg`, but I think we can just use a cast.

PR #156, opened 2019-03-23
Skip exit checks in unrolled loops

### LINKED ISSUE #127: Loop unrolling by HLS
Reading from this page on [how Vivado generates unroll loigc](https://japan.xilinx.com/html_docs/xilinx2017_4/sdaccel_doc/uyd1504034366571.html)

Excerpts:
> To unroll a loop completely, the loop bounds must be known at compile time. This is not required for partial unrolling.

So what happens in partial `unroll`s?
> Partial loop unrolling does not require N to be an integer factor of the maximum loop iteration count. Vivado HLS adds an exit check to ensure that partially unrolled loops are functionally identical to the original loop. For example, given the following code: 
```C
for(int i = 0; i < X; i++) {
  pragma HLS unroll factor=2
  a[i] = b[i] + c[i];
}
```
> turns into:
```C
for(int i = 0; i < X; i += 2) {
  a[i] = b[i] + c[i];
  if (i+1 >= X) break;
  a[i+1] = b[i+1] + c[i+1];
}
```

Woah, didn't know that. Turns out we don't need this exit check because we statically ensure all loops are aligned with their `unroll`s. Vivado gives us control over this: 
>  However, if you know that the specified unrolling factor, 2 in this example, is an integer factor of the maximum iteration count X, the `skip_exit_check option` lets you remove the exit check and associated logic.

Concretely, we need to do two things:

1. Do we want to support misaligned looks like these? (I think no, because we can always use the pattern from #72)

2. Start emitting `skip_exit_check` option for all our unrolls.

rachitnigam:
Emit `skip_exit_check` for unroll pragmas. According to the manual:
    
`skip_exit_check`: An optional keyword that applies only if partial unrolling is specified with `factor=`. The elimination of the exit check is dependent on whether the loop iteration count is known or unknown:
- Fixed (known) bounds: No exit condition check is performed if the iteration count is a multiple of the factor. If the iteration count is not an integer multiple of the factor, the tool:
   - Prevents unrolling.
   - Issues a warning that the exit check must be performed to proceed.
- Variable (unknown) bounds: The exit condition check is removed as requested. You must ensure that:
   - The variable bounds is an integer multiple of the specified unroll factor.
   - No exit check is in fact required.

Fixes #127.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -50,9 +50,3 @@ assembly := {
   getHeaders.value
   assembly.value
 }
-
-/* Running full test suite requires latest ./fuse build */
-test := {
-  assembly.value
-  (test in Test).value
-}
--- a/src/main/scala/GenerateExec.scala
+++ b/src/main/scala/GenerateExec.scala
@@ -57,16 +57,16 @@ object GenerateExec {
     }
 
     val CXX =
-      Seq("g++", "--std=c++11", "-Wall", "-I", headerLocation.toString) ++ compilerOpts
+      Seq("g++", "--std=c++14", "-Wall", "-I", headerLocation.toString) ++ compilerOpts
 
     val stderr = new StringBuilder
-    val logger = ProcessLogger(l => println(l),
+    val logger = ProcessLogger(l => scribe.info(l),
                                l => stderr ++= (l + "\n"))
 
     // Generate [[out]]. `!` is defined by sys.process:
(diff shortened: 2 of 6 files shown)


rachitnigam:
@sa2257 Make sure this is okay and ping me here if it isn't. Merging now.

PR #157, opened 2019-03-26
Buildbot run make

rachitnigam:
Two changes:
- Remove the old codepath that was used to just compile code sent in POST.
- Add `stage_make` for running make command on buildbot
  - The `make` parameter has to be explicitly passed to buildbot as a POST value. If missing, buildbot will not execute the make file.
  - Buildbout will run `make SDSFLAGS="..."` and assume that if the command terminates successfully, then the make command succeeded and it can continue from the HLS_FINISH state.

Testing this on buildbot.

The streaming log mentioned in #32 will be very useful now since synthesis commands take forever.


buildbot/buildbot/config_default.py
```diff
@@ -5,6 +5,8 @@
 SEASHELL_COMPILER = 'fuse'
 
 # A prefix command to use *before* the invocations of Xilinx tools.
+# NOTE(rachit): This is configured by a deployed instance of buildbot and
+# should be used whenever a command needs to be executed in the Xilinx container.
```
sampsyo: Or, a more universal way of putting it:

> For example, if your deployment needs to execute the Xilinx tools on a different machine or a different Docker container, a deployment could use this prefix to send these commands there.

buildbot/buildbot/state.py
```diff
@@ -1,4 +1,26 @@
...
+#         |                MAKE
+#       COMPILE              |
+#         |                  |
+#    COMPILE_FINISH          |
+#         |                MAKE_PROGRESS
+#        HLS                 |
+#         |                  |
+#      HLS_FINISH            |
+#         |                  |
+#        RUN-----------------+
```
sampsyo: Seems like this "edge" should connect to `HLS_FINISH`.

buildbot/buildbot/worker.py
```diff
@@ -230,7 +256,8 @@ def stage_seashell(db, config):
                     break
             else:
                 raise WorkError(
-                    'No hardware source file found. Expected a file with extension {} and basename not `main`.'.format(C_EXT)
+                    'No hardware source file found.' +
+                    ' Expected a file with extension {} and basename not `main`.'.format(C_EXT)
```
sampsyo: Python doesn't require `+` to concatenate string literals (to facilitate this kind of wrapping):

```
>>> ('a'
... 'b')
'ab'
```

So you can just leave that `+` off the previous line. (Similarly in the next diff hunk.)

buildbot/buildbot/worker.py
```diff
@@ -212,6 +223,21 @@ def stage_unpack(db, config):
                               os.path.join(task.code_dir, fn))
                 task.log('collapsed directory {}'.format(code_contents[0]))
 
+def stage_make(db, config):
+    """Work stage: run make command. Assumes that at the end of the make command,
+    work equivalent to the stage_hls is done, i.e., either estimation data has
+    been generated or a bitstrem has been generated.
```
sampsyo: typo: bitstrem

rachitnigam:
Things are looking good now. If this [test job](http://gorgonzola.cs.cornell.edu:8000/jobs/wNiAYmpTM_Q.html) goes through, then we're good to deploy.

I had to install `make` in the xilinx container.

buildbot/buildbot/worker.py
```diff
@@ -102,9 +102,10 @@ def run(self, cmd, log_stdout=True, timeout=60, cwd='', **kwargs):
 
 
 @contextmanager
-def work(db, old_state, temp_state, done_state):
+def work_with_func(db, old_state, temp_state, done_state_func):
```
sampsyo: I would like this a tiny bit more if this, instead of using a helper function, used a more Pythony way of letting the final argument be _either_ a state or a function that produces a state. Namely:

```py
def work(db, old_state, temp_state, done_state):
    # done_state is either a (constant) state or a function that
    # gives a final state for a given job.
    if isinstance(done_state, str):
        done_state = lambda _: done_state
```

rachitnigam: Fixed.

buildbot/buildbot/worker.py
rachitnigam: B U T   W H Y ?

Fixed.

buildbot/buildbot/state.py
rachitnigam: Fixed

buildbot/buildbot/config_default.py
rachitnigam: Fixed

buildbot/buildbot/worker.py
```diff
@@ -102,16 +102,20 @@ def run(self, cmd, log_stdout=True, timeout=60, cwd='', **kwargs):
...
 
 @contextmanager
-def work(db, old_state, temp_state, done_state):
+def work(db, old_state, temp_state, done_state_or_func):
     """A context manager for acquiring a job temporarily in an
     exclusive way to work on it. Produce a `JobTask`.
     Done state can either be a valid state string or a function that
     accepts a Task object and returns a valid state string.
     """
+    done_state = None
```
sampsyo: This line isn't necessary. :smiley: And a slightly better name for this might be `done_func`, since it's a function either way now.

rachitnigam: Is there no way to disable magic hoisting? I don't like variables being pulled up from `if` scope.

rachitnigam:
Works on buildbot. Ready to deploy.

buildbot/buildbot/worker.py
sampsyo: Only functions introduce scopes in Python, so you'd need to "thunk" the `if`.

rachitnigam: Facepalm. Okay, then let me keep the variable there for my own sanity. I've changed the name.

PR #158, opened 2019-03-27
buildbot: Tolerate bad job JSON files

sampsyo:
Raise an appropriate error when requesting a specific job, and do not include corrupted job in the list of "all" jobs. This should let the buildbot keep on truckin' without a mysterious 500 error when something bad happens.


rachitnigam:
Testing this with a rebase of #157 on buildbot.

rachitnigam:
Works correctly when a job is malformed.

PR #161, opened 2019-03-27
buildbot: Streaming log files

sampsyo:
As discussed in #32, the buildbot now uses plain, boring text files for its log (instead of keeping entries in its JSON file). This is sort of sad because we lose the ability to do more structured logging, such as keeping the timestamp as a plain integer, but the main upside is that invoked commands can "stream" their output to the log file. So it's possible to see the in-progress output from a command that has not finished yet.

I tested this locally and everything seems to be good, but the next step is to put this through its paces on Gorgonzola.


rachitnigam:
Seems to work on Gorgonzola, although the last 4 lines seem to be too few.

sampsyo:
Good point. Just a few lines made sense back when "lines" were entire command outputs. Fortunately, that's a configuration option now:
https://github.com/cucapra/seashell/pull/161/files#diff-bfda718cead065419d53e453bc2852c3R40

Maybe we should change that to ~50?

buildbot/buildbot/worker.py
```diff
@@ -110,7 +118,7 @@ def work(db, old_state, temp_state, done_state_or_func):
     """
     done_func = None
     if isinstance(done_state_or_func, str):
-        done_func = lambda _: done_state_or_func
+        done_func = lambda _: done_state_or_func  # noqa
```
rachitnigam: What's `noqa`?

buildbot/buildbot/templates/job.html
```diff
@@ -38,15 +38,7 @@ <h2>Job {{ job.name }}</h2>
...
 </p>
-<ul class="log">
-{% for time, text in job.log[-4:] %}
-    <li>
-        <time>{{ time | dt }}</time><br>
-        <pre>{{ text }}</pre>
-    </li>
-{% endfor %}
-</ul>
+<pre class="log">{{ log }}</pre>
```
rachitnigam: Is this just asking Flask to send the `log.txt` here? Could we instead use some JavaScript to continuously pull in the log and snip it to `LOG_PREVIEW_LINES` so we have a nice streaming log?

buildbot/buildbot/worker.py
sampsyo: That's an "ignore this line" directive for the Python style checker I use (and recommend!), [flake8](https://flake8.pycqa.org/en/latest/). The default settings warn on `foo = lambda: ...`, instead wanting you to write the equivalent but more verbose `def foo(): ...`.

buildbot/buildbot/templates/job.html
sampsyo: That string variable `log` is the trimmed version of the text file (which the server has already read and trimmed).

That fancy JavaScript version would be cool! But somewhat hard because it would need to use polling (or something complicated like inotify) because we don't know when a command will write more output to the file. And Server-Sent Events in the JavaScript to do the streaming. And async streaming on the server side, which is an awkward fit for Flask.

rachitnigam: I meant something even dumber like:
```
setInterval(updateLog, 10*1000)
```
where `updateLog` repeatedly sends a request to the server for `log.txt`. We can have some JavaScript to clear the timer when the state = done | failed.

sampsyo: Ah, I see. Yeah, we can try that too (but not in this PR, IMO—that’s still quite a change because it involves multiple JavaScripty requests instead of just this simple page formatting).

rachitnigam: Sounds good. This is ready to merge in that case.

PR #162, opened 2019-03-27
buildbot: Move timeouts to configuration

sampsyo:
[Based on #161, so the diff will look better after that's merged.]

The timeouts for compilation and synthesis are now in the buildbot's configuration, so they don't require editing hard-coded numbers to adjust anymore.


rachitnigam:
Should the base be `bb-logtxt`? If so, LGTM.

PR #163, opened 2019-03-29
Expose hw target and sdsflags as curl options on buildbot

rachitnigam:
As I was writing this, I realized that we can just set the platform and other flags in the Makefile itself. @sa2257 can just use the Makefiles while this is being tested (probably not today, I'm sick)

### DIFF
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -47,3 +47,6 @@ LOG_PREVIEW_LINES = 32
 # which has to be really long because synthesis is so slow.
 COMPILE_TIMEOUT = 120
 SYNTHESIS_TIMEOUT = 3000
+
+# Default Xilinx target platform.
+DEFAULT_PLATFORM = 'zed'
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -30,11 +30,13 @@ TEXT_EXTENSIONS = [
 ]
 
 # Configuration options allowed during job creation.
-CONFIG_OPTIONS = [
-    'skipseashell',
-    'estimate',
-    'make',
-]
+CONFIG_OPTIONS = {
+    'skipseashell': 'bool',
+    'estimate': 'bool',
+    'make': 'bool',
+    'sdsflags': 'str',
+    'platform': 'str',
+}
 
 # The number of (recent) lines of the log to show on job pages.
 LOG_PREVIEW_LINES = 32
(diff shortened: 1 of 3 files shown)


sa2257:
That's what I was saying yesterday. But the problem with that is we have to configure each Makefile. 

@rachitnigam, no worries! I'm not blocked by it, I'll be using local runs. Get well soon!

sampsyo:
OK, all done with a few suggestions! These are mostly just maintanability/usability things, except for https://github.com/cucapra/seashell/pull/163/commits/87b88f59c130582999fa7142a8b0b07b2fb1bc25 which addresses a problem that would prevent multiple arguments from working.

sa2257:
Out of curiosity, I read the values from configs without specifying a data type to them. Does specifying them make it easier to use? 
https://github.com/cucapra/seashell/commit/05624a14ebdb397472db89052b86f72454401f56
I also added `clock` which gives 4 frequencies to play around with, and might be useful.

sampsyo:
I see! Writing down the types just makes it easier to get the type of the field right without needing to add a *second* field called `foo_val` for every `foo`. (That is, it avoids the need for something like `hw_target_val` because the original boolean-ified `hw_target` won't be used (because the `bool(...)` version of it isn't useful).)

A `clock` option can't hurt, but it's also possible to specify by just supplying `-clkid` to `sdsflags`.

sampsyo:
Any objections to merging this now? (Before we fiddle too much with Makefile stuff as mentioned in #165.)

rachitnigam:
Sachille and I talked about this and we think it’s be best to keep the make stage dead simple.

Instead of adding SDSFLAGS for make, Sachille and I would prefer to have buildbot only call make and make ESTIMATE=1 and have it ignore other options. That way all SDS configuration lives in the makefile and we don’t have to rely on buildbot magic.

sa2257:
To be clear, I prefer the makefile to be even cleaner and Buildbot not assume anything. Buildbot should just run `make` and the makefile provider should ensure a self contained makefile without any Buildbot flags. (Or clearly defined set of flags)

sampsyo:
Sounds good. So the Buildbot doesn't provide SDSFLAGS at all—just higher-level options. That sounds just as reasonable.

rachitnigam:
Added the changes. Makefile stage only respects estimate flag and ignores the other flags.

sampsyo:
Looks good to me. I'm going to merge this now (and deploy it).

PR #168, opened 2019-04-06
Redex model for the core calculus

rachitnigam:


.gitignore
```diff
@@ -24,6 +24,10 @@ src/main/resources/headers/*.hpp
 ## Testing dir
 _test
 
+<<<<<<< HEAD
+=======
+*.o
+>>>>>>> 9718e6467c49c813e122acb886a07600bdfa4e57
```
sampsyo: Here are some merge markers! Conflicts are hard. :smile:

rachitnigam: R E B A S E

apurvakoti: Oops. Guilty as charged 😬

rachitnigam:
There are few things that this needs before we have something mergeable:

1. Refactor `Filament-ev` and `filament-->` to not have semantics for `par`. Since we want to play around with different implementations of `par`, these should be defined using language extensions.

2. Extend `Filament-ev` and `filament-->` to have a `stuck string` value (say into `Filament-ev-err` and `filament-with-err-->`. Instead of throwing racket errors in `lookup`, `add`, and `update`, the reduction rule should step to a stuck state with an error message. Obviously, this also means that the environment functions will be defined voer the `Filament-ev-err` language, not `Filament-ev`.

3. Add a test suite for this base language.

4. Create a new language `Filament-ev-err-par` that uses the current `par` semantics. Add tests if feasible. Skip if not.

@apurvakoti Do you have a sense of how long these tasks would take?

apurvakoti:
@rachitnigam 1. Simple enough - just remove par, make sure it still works
2. Shouldn't be too hard. Just add a new value and modify the semantics to step to this value on an error. The only ones we should need to modify are the ones that use the store at all (i.e. get, set)
3. An hour or so.
Minor name quibble: I feel like we should add name extensions the further we get from the full Filament, not closer; i.e. let the names "build up to" the final Filament language, rather than having increasingly unwieldy hyphens.

rachitnigam:
Okay. You can start working on these and ping me on Slack if you have questions. For the naming scheme, I don't care much what we call it. but note that there will be multiple `Filament` languages that we use explore different semantics for `par`. Building up to one language called `Filament` could be hard since we might want to add other stages in between.

Since Racket doesn't have restrictions on variable names, languages can have names like `Filament+err+par` to make it clearer. Similarly, theorems can have the form `reduce-Filament+err=reduce-Filament+par`. 

I suggest picking a scheme and sticking with it.

rachitnigam:
@apurvakoti any updates on this?

apurvakoti:
@rachitnigam Almost done. Using `apply-reduction-relation` to map over a list of expressions in the `par` form to get the result we discussed last time.

rachitnigam:
@apurvakoti is working on a separate, paper-based semantics for the language.

PR #180, opened 2019-04-22
Implement Views proposal

### LINKED ISSUE #124: Proposal: New view combinators
As we discussed today, and as a successor to #57's "shrink" and "flex," let's consider a new set of views. I'm calling these "combinators" because they are ways to take a memory and build a new memory out of it. (The output memory is, from the language's perspective, indistinguishable from a real, physical memory, so it can also be used in views.)

The goal is support the same patterns as in #57. As a quick summary, we're focused on convolution-style moving-window iteration (with overlap) and JPEG-style blocked iteration (without overlap). In the former case, we want to be able to unroll the inner loop (over the filter). In the latter case, we want to be able to unroll both loops.

So, let's build up a list of sufficient view combinators that get us there. For each combinator, we need four things: (1) the semantics, i.e., what values from the old memory show up in the new memory; (2) the banks to consume of the old memory; (3) the type of the new memory; and (4) some notion of the hardware cost.

Here's the list from today:

1. **Suffix**, a.k.a. cheap suffix. A slice expression like `a[i:]` in Python, where `i` must be a multiple of `a`'s banking factor. Consumes all of `a`. Produces a new, smaller memory with the same banking factor. The hardware is a cheap address offset.
2. **Rotation suffix**, a.k.a. expensive suffix. Also a slice like `a[i:]` in Python, but `i` is unconstrained. Also consumes all of `a` and produces a memory with the same banking factor. The hardware is a more costly bank rotation circuit.
3. **Prefix**. A slice like `a[:i]`. Consumes all of `a` and produces a smaller memory with the same banking factor. There is no hardware cost; this is just changes the type of the memory to allow access to a smaller part of it. For what it's worth, this combinator seems to argue in favor of *sound* bounds checking… otherwise, it would be possible to access the memory past the end of a prefix.
4. **Split**. Divide the banks of a memory into multiple, smaller memories. The semantics are kind of hard to explain, but the idea is that each of the several resulting memories represent discontiguous (strided) chunks of the original underlying array. The *total* banking factor of the resulting memories equals the banking factor of the original array, i.e., banks are neither created nor destroyed. All the banks of the original memory are consumed. The implied hardware seems to consist of almost nothing at all.
5. **Shrink**. Reduce the banking factor of an underlying array by an integer factor. The output memory is the same size as the input memory, just with a smaller banking factor—and the entire original array is consumed. The implied hardware is just enough fairly cheap muxing to allow N parallel PEs to access k*N banks by switching between bank sets in each iteration group.

We're getting closer! Some careful thought is still required to show precisely how to use these combinators to make our two notional loops we care about (convolution-like and JPEG-like) type-check.

rachitnigam:
Fixes #124


rachitnigam:
Before merging this in, open an issue about why views cannot be passed into functions as arguments.

PR #186, opened 2019-04-28
Add header generation

rachitnigam:
Adds new option `-h` that generates a header file if the given backend supports header generation.

@tedbauer can you test this for your use case before I merge this in?


rachitnigam:
Merging this seems the basics seem to work. Will fix bugs as they come up.

sa2257:
Does 'seem to work' mean 'generated a reasonable looking header file'? This doesn't also have an `include` keyword  which gets translated into `#include "fuse.h"` right? Put in another way, working header file doesn't mean it has a test does it?

rachitnigam:
> This doesn't also have an `include` keyword which gets translated into `#include "fuse.h"` right?

All `import`s in the original Fuse file will still get translated into `#include` equivalents if that's what you're asking about.

sa2257:
Ah great! So 'seem to work' means write a corresponding fuse header and write an `import` in `fuse` file and running the program? Awesome!

rachitnigam:
Hm, I'm confused by this question. I don't think it does what you're saying. 
You should try running `fuse -h src/test/should-run/matadd.fuse` to see what the output looks like.

PR #188, opened 2019-04-30
Explicit casting

### LINKED ISSUE #101: Operations between floats and ints
Is there any way to write something like this?
```
let x = 2.5;
let y = 1 + x;
```
This currently outputs:
```
[Type error] [2.11] + expected integers or floats, received: static(1) and float.
let z = 1 + x;
          ^
```

rachitnigam:
Fixes #101


src/main/scala/TypeCheck.scala
```diff
@@ -175,6 +176,15 @@ object TypeChecker {
     case EInt(v, _) => TStaticInt(v) -> env
     case EBool(_) => TBool() -> env
     case ERecLiteral(_) => throw RecLiteralNotInBinder(expr.pos)
+    case ECast(e, castType) => {
+      val (typ, nEnv) = checkE(e)
+      if (safeCast(typ, castType) == false) {
+        scribe.warn {
+          (s"Casting expression of type $typ to $castType which might not be safe", expr)
```
sampsyo: One tiny suggestion here: it seems like the rules for emitting a warning are trying to avoid classic sources of precision loss. So a more specific phrasing could say, "Casting X to Y may lose precision."

rachitnigam: Addressed in 7f375d0

PR #192, opened 2019-05-06
For loops with dynamic ranges

rachitnigam:
[WIP] Add support for `for` loops with dynamic ranges and steps.

@sa2257 @sampsyo @tedbauer comments for step syntax. Right now, it is:
```
for (let i = 0..10 step 2) ...
```

- Add support for loops with arbitrary expressions in place of start and end ranges.
- These loops cannot be unrolled.
- I'm worried that adding these is moving us towards the Vivado problem where certain magic constructs in the language can be unrolled whiles others can't be. Maybe it would be better to just use `while` loops as the un-`unroll`able loops and treating this as syntactic sugar.

### DIFF
--- a/src/main/scala/Errors.scala
+++ b/src/main/scala/Errors.scala
@@ -24,6 +24,9 @@ object Errors {
   case class UnrollRangeError(pos: Position, rSize: Int, uFactor: Int) extends TypeError(
     s"Cannot unroll range of size $rSize by factor $uFactor.", pos)
 
+  case class InvalidDynamicUnroll(pos: Position, uFactor: Int) extends TypeError(
+    s"Cannot unroll dynamic loop range with unroll factor $uFactor.", pos)
+
   case class InvalidIndex(id: Id, actual: Type) extends TypeError(
     s"Invalid indexing type for `$id'. Expected: TIndex or TSizedInt, actual: $actual", id.pos)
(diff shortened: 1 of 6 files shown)


tedbauer:
I think your suggested syntax is good because it's simple and clear.

Some random thoughts:
```
for i in [ 0..10 | step 2 ] { ...
```
```
for (let i = step 2: 0..10) { ...
```

rachitnigam:
Hm, I'm not happy with the current proposal. It conflicts with the current syntax making parsing gnarly and it isn't clean when a `for` loop can be unrolled and when it can't.

sampsyo:
I don't mind the `step` syntax all by itself! Reminds me of Apple BASIC. :smiley:
https://rosettacode.org/wiki/Loops/For_with_a_specified_step#Applesoft_BASIC

However, this is an instance where we could consider embracing the all-consuming **everything is a view** philosophy. That is, to iterate over this array, you could take a new kind of view that strides over an underlying array. You can imagine the hardware gadget that would do the index adjustment: something along the lines of `i => i*s` where `s` is the step size. This would be cheap as long as `s` divides the banking factor—and it also seems straightforward to compute the new memory type of the strided view. Then, the loop would just iterate over the view instead of the underlying array.

PR #195, opened 2019-05-14
Support for array literals

### LINKED ISSUE #181: Array creation construct
Moving issue about possibly adding a Fuse construct from fuse-benchmarks here. Old issue: cucapra/fuse-benchmarks#40

rachitnigam:
Syntax changes:
- Array literals of the form `{1, 2, 3, 4}`.
  - Only only with an explicit let binder.
  - Only single dimensional arrays supported.

Fixes #181. Subsumes #182.


docs/binders.md
```diff
@@ -28,8 +29,16 @@ can be represented as the explicit type.
...
+> The right hand side has the type `bit<1>` which can be zero extended to be
+> represented using 10 bits.
+
+The final form allows defining uninitialized variables. Reading from uninitialized
+memories is undefined behaviour.
+
+```
+let x: bit<10>;
+let y = x; // undefined behavior
+```
```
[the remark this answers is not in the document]
rachitnigam: Ah, I meant to say that the returned value is an undefined value. I didn't realize the distinction between the two.

src/main/scala/TypeCheck.scala
```diff
@@ -175,7 +175,8 @@ object TypeChecker {
     case EFloat(_) => TFloat() -> env
     case EInt(v, _) => TStaticInt(v) -> env
     case EBool(_) => TBool() -> env
-    case ERecLiteral(_) => throw RecLiteralNotInBinder(expr.pos)
+    case ERecLiteral(_) => throw NotInBinder(expr.pos, "Record Literal")
+    case EArrLiteral(_) => throw NotInBinder(expr.pos, "Array Literal")
```
rachitnigam: @sampsyo wirt your first concern: We do a simple well-formedness check here to make sure that array literals don't exist anywhere but inside Let binders. 

This is hack which underlines two issues: (1) Well-formedness checks should not occur inside the type checker, they should be separate. (2) The AST is more stratified that Syntax.scala lets on. There should be some way to making this more explicit.

sampsyo: Got it; thanks. FWIW, although it's totally fine to rule them out here for now, I think there's no harm in record literals occurring anywhere (following the signal-vs-memory distinction mentioned below).

rachitnigam:
I agree the solution is less than ideal with local memories. However, I don't have a good solution beyond just adding more syntax forms.

If the current solution is reasonable, we should merge this and come back to this later on.

sampsyo:
Yep, totally agreed! Let’s merge this and address the philosophical misgivings in the future.

PR #196, opened 2019-05-15
Gadget checker

### LINKED ISSUE #193: Views and underlying array being used together
More views complications. Consider this program:

```
view v_a = a[i! :];
---
v_a[10]; a[20];
```

Fuse will happily mark usage of the second access to `a` as correct because it considers `v_a` to be "just an array". Looks like view creation needs some more complicated logic for safety...

rachitnigam:
Fixes #193


src/test/scala/TypeCheckerSpec.scala
```diff
@@ -203,12 +184,10 @@ class TypeCheckerSpec extends FunSpec {
       }
     }
     it("adding static ints does NOT perform type level computation") {
-      val e1 = typeCheck("let x = 1; let y = 2; let z = x + y;")
-      assert(e1("z") === TSizedInt(2))
+      typeCheck("let x = 1; let y = 2; let z = x + y;")
```
sampsyo: Stating the obvious here, but this test (like the ones below) is now a little weaker: it doesn't check the deduced type for `z`.

rachitnigam:
Added explanation comments. Ready to merge. @sa2257 and @sgpthomas should read this at some point and ask questions

PR #198, opened 2019-05-29
Separation2

sgpthomas:
I reorganized the repository into folders. This is based on the work that we did in the separate branch but I redid it on master. I also made sure that `common` is self contained and doesn't depend on code outside of `common`. In order to do this, I had to duplicate `Configuration` case class. I split it into `Configuration` and `CmdlineConfig`. They are the same except that `CmdlineConfig` has fields for `backend` and `mode`. Furthermore, I put all the capability checking stuff into the same folder as the typechecking stuff. I don't know if they should be further organized into separate folders.


sgpthomas:
Finally returning back to this. @rachitnigam I might be misunderstanding something but I don't think Scala supports having a single file in different namespaces. Scala does support this:
```scala
package org.myproject
package tests
```
However, this doesn't mean that the file is in two different namespaces. But rather that it is in the namespace org.myproject.tests. The reason to write this with 2 package statements rather than 1 is so that both things from `org.myproject` and `org.myproject.tests` are available in the file.
Here's a blog post about this: [https://www.artima.com/scalazine/articles/chained_package_clauses_in_scala.html](https://www.artima.com/scalazine/articles/chained_package_clauses_in_scala.html). This is from scala 2.8 so it's possible things have changed since then but I haven't found anything more recent.

rachitnigam:
Oh interesting! That setup does make more sense than a single file belonging in multiple packages.

sampsyo:
Yo; what can I do to help get this ready to merge? Looks like the test runner just needs to get synced up (according to Circle), but then are we good to go? I can take a crack at that.

sgpthomas:
yes that should be it. Also could you verify the change that I made to get rid of two configs? To summarize, I just leave the backend as a string a little bit longer. I convert it to a proper backend right before we are going to use the backend rather than at command line parse time

sampsyo:
OK, I got the test runner working again (and Circle seems to agree).

I liked the "move backend instantiation until later" thing, @sgpthomas, but I tweaked it a little: instead of literally being strings, I made a little "enum" (not really an enum) describing the possible backends. Should be a tad more type safe.

If this looks OK to @rachitnigam, let's merge this! Then @KarlMobed's new backend can hook itself in using the new style.

sampsyo:
I even rebased to resolve conflicts with master. I hope you're happy, @rachitnigam. 😆

PR #200, opened 2019-05-30
buildbot: Copy datafiles to FPGA

sa2257:
Copying data to sd_card directory before it's copied over to FPGA. [test](http://gorgonzola.cs.cornell.edu:8000/jobs/a0YznOGWOVM.html)

Need to handle case where there's no datafiles to copy. [issue](http://gorgonzola.cs.cornell.edu:8000/jobs/eR1B6evg5Hg.html)

This PR is branched from #199

### DIFF
--- a/buildbot/buildbot/worker.py
+++ b/buildbot/buildbot/worker.py
@@ -17,6 +17,7 @@ HOST_O = 'main.o'  # The .o file for host code.
 # For executing on a Xilinx Zynq board.
 ZYNQ_SSH_PREFIX = ['sshpass', '-p', 'root']  # Provide Zynq SSH password.
 ZYNQ_HOST = 'zb1'
+ZYNQ_DEST_DIR = '/mnt'
 ZYNQ_REBOOT_DELAY = 40
 
 
@@ -364,7 +365,7 @@ def stage_fpga_execute(db, config):
         # Zynq board.
         bin_dir = os.path.join(task.code_dir, 'sd_card')
         bin_files = [os.path.join(bin_dir, f) for f in os.listdir(bin_dir)]
-        dest = ZYNQ_HOST + ':/mnt'
+        dest = '{}:{}'.format(ZYNQ_HOST, ZYNQ_DEST_DIR)
         task.run(
             ZYNQ_SSH_PREFIX + ['scp', '-r'] + bin_files + [dest],
             timeout=1200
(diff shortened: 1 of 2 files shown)


buildbot/buildbot/worker.py
```diff
@@ -360,11 +361,18 @@ def stage_fpga_execute(db, config):
             task.log('skipping FPGA execution stage')
             return
 
+        # Copy datafiles to executable dir.
+        data_files = [os.path.join(task.code_dir, f) for f in os.listdir(task.code_dir) if f.endswith('.data')]
+        dest = os.path.join(task.code_dir, 'sd_card')
+        task.run(
+            ['cp'] + data_files + [dest]
+        )
```
sampsyo: It seems like it would be a good idea to skip this if `data_files` is empty (put it under `if data_files:`). Perhaps a safer way to do this would be to use [shutil.copy](https://docs.python.org/3.7/library/shutil.html#shutil.copy) instead of `cp`.

sa2257: condition check is included now. I didn't use shutil.copy yet.

sa2257: It seems shutil.copy permits copying files one at a time. Unless there's an additional benefit, I'll mark the PR ready.

sampsyo:
Aside from that suggestion, I had also left a different comment above:

> Our Makefile is already responsible for copying data files into `sd_card`, so this should probably go on the non-Makefile path, i.e., in the `stage_hls` function.

Also, I notice that the most recent version includes `if f is not None`, which I think is unnecessary? The program would crash anyway when calling a method on `f` if it were None. And I believe that `listdir` does not return Nones.

sa2257:
> Aside from that suggestion, I had also left a different comment above:
> 
> > Our Makefile is already responsible for copying data files into `sd_card`, so this should probably go on the non-Makefile path, i.e., in the `stage_hls` function.
> 

Oh indeed! I'll move that to `stage_hls`.

> Also, I notice that the most recent version includes `if f is not None`, which I think is unnecessary? The program would crash anyway when calling a method on `f` if it were None. And I believe that `listdir` does not return Nones.

I think it's no longer necessary. I'll remove that too. Thanks!

sampsyo:
OK, looks good! Let’s merge this. But maybe it would be worth documenting somewhere that, to have your files copied, they have to have the `.data` extension?

sa2257:
> OK, looks good! Let’s merge this. But maybe it would be worth documenting somewhere that, to have your files copied, they have to have the `.data` extension?

Great idea! Will this line on buildbot readme do?

sampsyo:
Sounds good!

PR #201, opened 2019-06-14
Update overview.md

KarlMobed:
Tried to replace and clarify a made up work "realizability"

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -4,4 +4,4 @@ title: Overview
 ---
 
 Fuse is a programming language for designing hardware accelerators.
-It provides abstractions that guarantee hardware realizability after type checking.
+It provides abstractions that guarantee the same behavior as hardware after type checking.


sa2257:
Also to clarify, "hardware realization" is a phrase used frequently in hardware design (Searching this in Google/ Google Scholar will verify that).

KarlMobed:
Interesting information to learn.  I google or ask about cases like this in the future instead of siding with spellcheck and assuming a misspelling

sa2257:
Essentially our abstractions should guarantee two things,
- They should be synthesizable in hardware (HLS tools by the way, for the most part guarantees this)
- They should be hardware friendly, that's to say, the designs should not be a naive emulation of C behavior and designed to leverage underlying hardware.

rachitnigam:
Closing since there doesn't seem anything to do.

PR #203, opened 2019-06-18
Array literals cannot be used across --- boundaries

rachitnigam:
Only contains the failing test. This is because the environment containing physical resources are not merged after `---`.

### DIFF
--- a/src/test/scala/TypeCheckerSpec.scala
+++ b/src/test/scala/TypeCheckerSpec.scala
@@ -1163,6 +1163,14 @@ class TypeCheckerSpec extends FunSpec {
         x[0] = false
         """ )
     }
+    it("can be used across --- boundaries") {
+      typeCheck("""
+        let x: bool[3] = {true, false, true};
+        x[1];
+        ---
+        x[0] := false
+        """ )
+    }
   }
 
   describe("Indexing with dynamic (sized) var") {


rachitnigam:
Given this example:
```
let a: float[10];
a[0];
---
a[0];
```

The issue is the the scope of the physical resource (not the type or the gadget) defined by the `let a ...` ends at the `---`. The quickfix is turning the above into:

```
let a: float[10];
{
  a[0];
  ---
  a[0];
}
```

Fixing the other issue of trying extend the scope of the physical resource is harder because the left side of `---` consumes some banks and we need to replicate the resource but not the consumed banks.

@sampsyo what syntax form feels more reasonable? The issue occurs because of the same reasons that we discussed in #195 where `let` bound arrays aren't actually local variables and need to be treated specially because they imply creation of special memories.

sampsyo:
I'm of two minds, because the second code block above really is more "logical" than the first—you'd want to explicitly make `a` "fan out" to the two stages. But the former is, of course, more like how normal imperative programming works—sequential composition with `;` doesn't make your variables go away. It's like the pre-C99 days when you had to declare all your variables at the top of the function.

This may be a case where we want the former to be sugar for the latter. But enforcing the latter seems harmless for now.

ViviYe:
`while (k <= stop) {
  let temp_j: bit<32>  = temp[j];
  ---
  let temp_i: bit<32>  = temp[i];
  if(tmp_j < tmp_i) {
            a[k] := tmp_j;
            j := j-1;
        } else {
            a[k] := tmp_i;
            i := i+1;
        }
  
}`
In this piece of code, if no curly braces added, temp_j can't be accessed by the if statements below.
However, it seems like  --- is not allowed right after {. The compiler won't accept 
let temp_j: bit<32>  = temp[j];{
  --- 
...
}
Is there a good way to get around this situation?

ViviYe:
In this piece of code,

```
while (k <= stop) {
  let temp_j: bit<32>  = temp[j];
  ---
  let temp_i: bit<32>  = temp[i];
  if(tmp_j < tmp_i) {
            a[k] := tmp_j;
            j := j-1;
        } else {
            a[k] := tmp_i;
            i := i+1;
        }
}
```

for temp_j to be accessed by the code below, curly braces are needed. However, --- is not allowed right after curly braces.
Is there a good way to get around this situation?

sampsyo:
The tricky part of this example is that there *needs* to be a timestep boundary (i.e., `---`) between the two `let`s because `temp` is being reused there. But that means that we can't do the trick @rachitnigam mentioned above because it would have to look like this:

```
let temp_j = temp[j];
{
    ---
    let temp_i = temp[i];
    {
       (other stuff here)
    }
}
```

As @ViviYe mentioned, this doesn't even parse. (@ViviYe, my silly suggestion for now was just to put a useless line of code right before the `---` in the listing above just to get the thing to parse.)

I think what we really want is something like this:

```
let temp_i;
let temp_j;
{
    temp_i = temp[I];
    ---
    temp_j = temp[j];
    ---
    (other stuff here)
}
```

…that is, the name declarations come from the outer scope, but the assignments happen in sequence. I don't know if this is allowed precisely (probably not?), but @ViviYe, you can do this by assigning dummy values (i.e., 0) at the initial declaration points and then overwriting that with the real values.

---

But this does make me think more deeply about what's going on here, and it's strange that—in these programs—the declarations are "concurrent" with the writes to the accesses to the variable. That is, there may be time steps inside the curly braces, but that entire sequence is in a _parallel_ composition with the `let`. Seems odd.

ViviYe:
```
while (k <= stop) {
  let temp_j: bit<32>  = temp[j];
  {
  let x = 0;
  ---
  let temp_i: bit<32>  = temp[i];
  if(tmp_j < tmp_i) {
            a[k] := tmp_j;
            j := j-1;
        } else {
            a[k] := tmp_i;
            i := i+1;
        }
  
   }
  }
```
I tried to insert a useless line of code before ---
However the compiler is still yelling at me : 

```
[31.31] Bank 0 in dimension 1 of physical resource `temp' already consumed.
  let temp_i: bit<32>  = temp[i];
                              ^

[27.31] Last consume happened here:
  let temp_j: bit<32>  = temp[j];
                              ^
```
I couldn't figure out why this is the case, since I already had a --- seperating the two statements that access temp.

sampsyo:
I think it's the problem I mentioned above: the parallel-composition `;` means that the two things on either side might run in parallel. Even though there is a `---` within the block on the right-hand side, that's still running in parallel with the `let` on the left-hand side.

How about the second approach I mentioned above?

ViviYe:
```
while (k <= stop) {
  let temp_j: bit<32>  = 0;
  {
  temp_j := temp[j];
  ---
  let temp_i: bit<32>  = temp[i];
  if(tmp_j < tmp_i) {
            a[k] := tmp_j;
            j := j-1;
        } else {
            a[k] := tmp_i;
            i := i+1;
        }
  
   }
  }
```
I am not sure if I understood the second approach correctly. I tried this but now the error is 
```
[Type error] [32.6] `tmp_j' is not bound in scope.
  if(tmp_j < tmp_i) {
     ^
```
I assume the if statement should be in the scope?

ViviYe:
I Also tried moving both declaration of temp_j and temp_I outside the curly braces and still got the same error message. : (

sampsyo:
Huh! That is indeed what I meant, and this one has me stumped.

I think the next step, @ViviYe, might be to try "minimizing" this example to come up with a very short program that triggers the same surprising behavior. I usually do this by making a copy of the code and going through an iterative process like this:

1. Compile it to make sure the error happens.
2. Delete some code.
3. Compile it again to see if the error still happens.
    - If yes, go to 2 (deleting even more code).
    - If no, restore the code I just deleted and go to 2 (deleting different code this time).
    - If no and there doesn't seem to be anything left to delete, you're done and have a minimal example!

I think Real Software Engineers™ call this [delta debugging](https://en.wikipedia.org/wiki/Delta_debugging).

ViviYe:
Through minimizing it I found out what is wrong. I had a typo! Thank you so much for helping me debug!!

rachitnigam:
Just to clarify from this conversation. The restriction above **only** applied to `let` bound array literals right now. Normal literals can cross `---` just fine.

PR #204, opened 2019-06-19
Buildbot features

sa2257:
Adding features to Buildbot
- hw_basename for make file runs
   Not entirely satisfied. Names update only once the job is done.
- accepting directive files
   Make file flow provides make command with file name, regular flow generates the command.

### DIFF
--- a/buildbot/README.md
+++ b/buildbot/README.md
@@ -69,6 +69,8 @@ You can also specify job configuration options as further POST parameters:
 - `skipseashell`, which lets you supply plain HLS C code as input.
 - `estimate`, to use the Xilinx toolchain's resource estimation facility. The job will skip synthesis and execution on the FPGA.
 - `make`, to use a Makefile instead of the built-in compilation workflow (see below).
+- `hwname`, which lets you provide a name for the job during makefile flow.
+- `directives`, which let's you provide the name of a file with a set of directives (pragmas). If you want to force default directive files to be ignored you can use this as `directives=0`.
 
 Use `-F <option>=1` to enable these options with `curl`.
 
@@ -96,5 +98,6 @@ For estimation, buildbot supplies flags for `sds++` using the `SDSFLAGS` variabl
 %.o: %.c
     sds++ $(SDSFLAGS) $< -o $@
 ```
+For configurations provided above, buildbot also relies on following variables, `ESTIMATE`, `DIRECTIVES`. Make sure these variables behave according to the configuration description in your makefile.
 
 [curl]: https://curl.haxx.se
--- a/buildbot/README.md
+++ b/buildbot/README.md
@@ -70,7 +70,7 @@ You can also specify job configuration options as further POST parameters:
 - `estimate`, to use the Xilinx toolchain's resource estimation facility. The job will skip synthesis and execution on the FPGA.
 - `make`, to use a Makefile instead of the built-in compilation workflow (see below).
 - `hwname`, which lets you provide a name for the job during makefile flow.
-- `directives`, which let's you provide the name of a file with a set of directives (pragmas). If you want to force default directive files to be ignored you can use this as `directives=0`.
+- `directives`, which lets you provide the name of a file with a set of directives (pragmas).
 
 Use `-F <option>=1` to enable these options with `curl`.
 
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -45,7 +45,7 @@ CONFIG_OPTIONS = {
 EXECUTABLE_NAME = 'sdsoc'
 
 # Default directives file is empty.
-DEFAULT_DIRECTIVES = ' '
+DEFAULT_DIRECTIVES = ''
 
 # The number of (recent) lines of the log to show on job pages.
 LOG_PREVIEW_LINES = 32
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -35,6 +35,7 @@ CONFIG_OPTIONS = {
     'skipseashell': bool,
     'estimate': bool,
     'make': bool,
+    'kernel': str,
     'sdsflags': str,
     'platform': str,
 }
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -36,16 +36,13 @@ CONFIG_OPTIONS = {
     'estimate': bool,
     'directives': str,
     'make': bool,
-    'kernel': str,
+    'hwname': str,
     'sdsflags': str,
     'platform': str,
 }
 
 # The name to use for compiled executables.
 EXECUTABLE_NAME = 'sdsoc'
-
-# Default directives file is empty.
-DEFAULT_DIRECTIVES = ''
 
 # The number of (recent) lines of the log to show on job pages.
 LOG_PREVIEW_LINES = 32
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -34,6 +34,7 @@ TEXT_EXTENSIONS = [
 CONFIG_OPTIONS = {
     'skipseashell': bool,
     'estimate': bool,
+    'directives': str,
     'make': bool,
     'kernel': str,
     'sdsflags': str,
@@ -42,6 +43,9 @@ CONFIG_OPTIONS = {
 
 # The name to use for compiled executables.
 EXECUTABLE_NAME = 'sdsoc'
+
+# Default directives file is empty.
+DEFAULT_PLATFORM = ' '
 
 # The number of (recent) lines of the log to show on job pages.
 LOG_PREVIEW_LINES = 32
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -45,7 +45,7 @@ CONFIG_OPTIONS = {
 EXECUTABLE_NAME = 'sdsoc'
 
 # Default directives file is empty.
-DEFAULT_PLATFORM = ' '
+DEFAULT_DIRECTIVES = ' '
 
 # The number of (recent) lines of the log to show on job pages.
 LOG_PREVIEW_LINES = 32
--- a/buildbot/buildbot/templates/joblist.html
+++ b/buildbot/buildbot/templates/joblist.html
@@ -42,7 +42,7 @@ Or see the <a href="{{ url_for('live_html') }}">live interface</a>.
                     {{ job.name }}
                 </a>
             </td>
-            <td>{{ job.hwname | default(job.hw_basename) }}</td>
+            <td>{{ job.config.hwname | default(job.hw_basename) }}</td>
             <td>{{ job.started | dt }}</td>
             <td class="{{ job.state }} status">{{ status_strings[job.state] }}</td>
         </tr>
--- a/buildbot/buildbot/templates/joblist.html
+++ b/buildbot/buildbot/templates/joblist.html
@@ -42,7 +42,7 @@ Or see the <a href="{{ url_for('live_html') }}">live interface</a>.
                     {{ job.name }}
                 </a>
             </td>
-            <td>{{ job.hw_basename }}</td>
+            <td>{{ job.hwname }}</td>
             <td>{{ job.started | dt }}</td>
             <td class="{{ job.state }} status">{{ status_strings[job.state] }}</td>
         </tr>
--- a/buildbot/buildbot/templates/joblist.html
+++ b/buildbot/buildbot/templates/joblist.html
@@ -42,7 +42,7 @@ Or see the <a href="{{ url_for('live_html') }}">live interface</a>.
                     {{ job.name }}
                 </a>
             </td>
-            <td>{{ job.config.hwname or job.hw_basename }}</td>
+            <td>{{ job.hwname | default(job.hw_basename) }}</td>
             <td>{{ job.started | dt }}</td>
             <td class="{{ job.state }} status">{{ status_strings[job.state] }}</td>
         </tr>
--- a/buildbot/buildbot/templates/joblist.html
+++ b/buildbot/buildbot/templates/joblist.html
@@ -42,7 +42,7 @@ Or see the <a href="{{ url_for('live_html') }}">live interface</a>.
                     {{ job.name }}
                 </a>
             </td>
-            <td>{{ job.hwname }}</td>
+            <td>{{ job.config.hwname or job.hw_basename }}</td>
             <td>{{ job.started | dt }}</td>
             <td class="{{ job.state }} status">{{ status_strings[job.state] }}</td>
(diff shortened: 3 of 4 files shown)


sa2257:
Note to self- There's a bug in the regular flow.

buildbot/buildbot/config_default.py
```diff
@@ -34,14 +34,19 @@
...
+    'kernel': str,
     'sdsflags': str,
     'platform': str,
 }
 
 # The name to use for compiled executables.
 EXECUTABLE_NAME = 'sdsoc'
 
+# Default directives file is empty.
+DEFAULT_DIRECTIVES = ''
```
sampsyo: I don't think we need this—this is the only sensible default for `directives`; it wouldn't make sense for this to be nonempty.

buildbot/buildbot/worker.py
```diff
@@ -154,6 +154,8 @@ def _task_config(task, config):
     task['sdsflags'] = flags
     task['estimate'] = est
 
+    task['directives'] = task['config'].get('directives') or \
+        config['DEFAULT_DIRECTIVES']
```
sampsyo: Accordingly with the above, I think it's fine for this to just leave off the `or` part and instead pass `''` as the second parameter to `.get()`, which will make this an empty string by default.

buildbot/buildbot/worker.py
```diff
@@ -226,6 +230,7 @@ def stage_make(db, config):
             prefix + [
                 'make',
                 'ESTIMATE={}'.format(task['estimate']),
+                'DIRECTIVES={}'.format(task['directives']),
```
sampsyo: I assume you probably want to not pass anything at all to the Makefile when `task['directives']` is empty? Otherwise, this will overwrite a `DIRECTIVES` that's defined in the Makefile already, and you probably want that to be filled with the "correct" directives name by default.

buildbot/buildbot/worker.py
```diff
@@ -285,17 +290,27 @@ def stage_seashell(db, config):
...
+        ]
+    else:
+        return prefix + [
+            'sds++',
+            '-sds-pf', task['platform'],
+            '-sds-hw', func_hw, c_hw, '-hls-tcl', task['directives'], '-sds-end',
+            '-clkid', '3',
+            '-poll-mode', '1',
+            '-verbose', '-Wall', '-O3',
+        ]
```
sampsyo: Let's avoid duplication here by building the constant part of the list and adding in only the necessary parts when there are directives.

buildbot/buildbot/worker.py
```diff
@@ -315,10 +330,11 @@ def stage_hls(db, config):
         hw_basename, hw_c, hw_o = _hw_filenames(task)
         _task_config(task, config)
         flags = shlex.split(task['sdsflags'])
+        sds_cmd = _sds_cmd(prefix, hw_basename, hw_c, task)
```
sampsyo: If `task` is going to be an argument to `_sds_cmd`, then the `hw_basename` and `hw_c` parameters can also just come straight from `task` (and get looked up within there). And we should probably also just append the flags in there, right?

buildbot/buildbot/worker.py
```diff
@@ -217,6 +219,8 @@ def stage_make(db, config):
     with work(db, state.MAKE, state.MAKE_PROGRESS, state.HLS_FINISH) as task:
         _task_config(task, config)
 
+        task['hw_basename'] = task['config'].get('kernel')
```
sampsyo: Rather than "overloading" this field's meaning, which was originally meant to control the Makefile-free code generation step, let's instead just display `task['config']['kernel']` in the UI directly. Here's the relevant line of code in the UI template:
https://github.com/cucapra/seashell/blob/79b297306035a24d3af9913cd34f8ad583a529bb/buildbot/buildbot/templates/joblist.html#L45

buildbot/buildbot/config_default.py
```diff
@@ -34,14 +34,19 @@
 CONFIG_OPTIONS = {
     'skipseashell': bool,
     'estimate': bool,
+    'directives': str,
     'make': bool,
+    'kernel': str,
```
sampsyo: Maybe "name" would be a better name for this field than "kernel"? It's just some identifying string meant to be helpful to humans, without any prescribed deeper meaning than that…

buildbot/buildbot/worker.py
sa2257: That's a great point. I didn't think of that. 
So, I'm going to change it to the following,
One can submit a directive file to buildbot, buildbot will run the make command with `DIRECTIVES` set to the file name. Otherwise, it'll run it as before.
I might still be missing something from our chat on Friday.
Is there a case to say `DIRECTIVES=0` or `1`?

buildbot/buildbot/worker.py
sa2257: Yes, thanks!

buildbot/buildbot/worker.py
sampsyo: Hmm... that sounds a little more complicated than I was imagining. Here’s what I was suggesting:

- The Makefile has `DIRECTIVES := foo` or whatever in it already.
- The way you use this is that just typing `make` without any arguments at all will use directives from foo. If you want to tell the thing not to use directives, do `DIRECTIVES=` on the command line.
- Buildbot passes nothing by default. Only when `DIRECTIVES` is set to something does it pass something to `make` on the command line.

Easy! No need for automatic file detection.

sa2257: Ah yes, that sounds great! Thanks!

buildbot/buildbot/worker.py
sa2257: I'm not quite sure whether I resolved this in the cleanest way.

buildbot/buildbot/worker.py
```diff
@@ -285,17 +294,27 @@ def stage_seashell(db, config):
...
+            '-sds-pf', task['platform'],
+        ]
+    sds_cmd_tail = [ 
+            '-clkid', '3',
+            '-poll-mode', '1',
+            '-verbose', '-Wall', '-O3',
+        ]
+    if task['config'].get('directives'):
+        return sds_cmd_head + [
+                '-sds-hw', func_hw, c_hw, '-hls-tcl', task['config'].get('directives'), '-sds-end',
```
sampsyo: As above, you don’t need `get` here.

buildbot/README.md
```diff
@@ -69,6 +69,8 @@ You can also specify job configuration options as further POST parameters:
 - `skipseashell`, which lets you supply plain HLS C code as input.
 - `estimate`, to use the Xilinx toolchain's resource estimation facility. The job will skip synthesis and execution on the FPGA.
 - `make`, to use a Makefile instead of the built-in compilation workflow (see below).
+- `hwname`, which lets you provide a name for the job during makefile flow.
+- `directives`, which let's you provide the name of a file with a set of directives (pragmas). If you want to force default directive files to be ignored you can use this as `directives=0`.
```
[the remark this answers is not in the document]
sampsyo: Also, Buildbot (with this change) doesn’t seem to assign any particular semantics to `directives=0`; it just blindly passes it along to the Makefile (and doesn’t do anything special with it in the non-Makefile path). So maybe this description would be better left to the Makefile, which actually interprets the “zero” as something special. Or at least clarifying that this is not the Buildbot’s territory.

buildbot/buildbot/templates/joblist.html
```diff
@@ -42,7 +42,7 @@ <h2>Jobs</h2>
                     {{ job.name }}
                 </a>
             </td>
-            <td>{{ job.hw_basename }}</td>
+            <td>{{ job.hwname }}</td>
```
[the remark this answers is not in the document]
sa2257: No reason. This is certainly a better way to do it.

buildbot/README.md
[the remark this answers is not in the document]
sa2257: I see, it doesn't indeed do anything for non-Makefile path. Why added it was to say that `DIRECTIVES=` or `DIRECTIVES= " "` wouldn't work as they'd skip the `get('directives')` and use the default file, in makefile path. I assume `0` worked because `sds++ -hls-tcl 0` also worked as intended.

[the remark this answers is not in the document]
sampsyo: Another alternative, FWIW, is to *not* do `.get()` stuff and make the presence of an empty value (`-F directives=`) actually different from leaving it off altogether.

buildbot/buildbot/templates/joblist.html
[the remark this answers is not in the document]
sa2257: The change is giving me some trouble. I need to debug it.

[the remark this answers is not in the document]
sa2257: Using `job.config.hwname` gives me an error when loading the webpage
```
Internal Server Error
The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
```

docker logs have
```
File "/seashell/buildbot/buildbot/templates/joblist.html", line 45, in block "content"
<td>{{ job.config.hwname or job.hw_basename }}</td>
File "/root/.local/share/virtualenvs/buildbot-mKkm-6u_/lib/python3.6/site-packages/jinja2/environment.py", line 430, in getattr
return getattr(obj, attribute)
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'config'
```

I'm somewhat stuck in my debugging, any pointers appreciated.

To do
- [ ] Remove https://github.com/cucapra/seashell/blob/e4a68b555aa58eea73da789fdc7c3704d06da237/buildbot/buildbot/worker.py#L221, it's currently used for debugging
- [ ] Update docs with the suggested change above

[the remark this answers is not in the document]
sampsyo: This should do it: `hw.config.jobname | default(job.hw_basename)`

The docs should help:
http://jinja.pocoo.org/docs/2.10/templates/#default

sa2257:
Bug found in [job](http://gorgonzola.cs.cornell.edu:8000/jobs/rbtFF0Fmp1o.html)

sa2257:
Bug fixed, but 
> Another alternative, FWIW, is to _not_ do `.get()` stuff and make the presence of an empty value (`-F directives=`) actually different from leaving it off altogether.
is not achieved yet. 

Replacing `task['hwname'] = task['config'].get('hwname')` with displaying `task['config'].get('hwname')` on job page is not done either.

sampsyo:
@sa2257, I know you're out for a few days, but when you're back, please see the above commit. I just tried the `| default(...)` version I wrote above, and it seems to work.

Did you try this before and get some kind of error? If so, what error? I think you never showed me which error it was. It would be surprising if exactly the same code gave you an error I can't reproduce.

sa2257:
> @sa2257, I know you're out for a few days, but when you're back, please see the above commit. I just tried the `| default(...)` version I wrote above, and it seems to work.
> 
> Did you try this before and get some kind of error? If so, what error? I think you never showed me which error it was. It would be surprising if exactly the same code gave you an error I can't reproduce.

The error seems to be due to old jobs without config. Old jobs were selectively removed but there are some jobs without an `info.json` which escaped the clean up. Therefore, temporarily all jobs prior jobs are moved to `job_archive` folder, to gradually move the valid jobs.

buildbot/buildbot/worker.py
```diff
@@ -285,17 +291,28 @@ def stage_seashell(db, config):
...
-def _sds_cmd(prefix, func_hw, c_hw, platform):
+def _sds_cmd(prefix, task):
     """Make a sds++ command with all our standard arguments.
     """
-    return prefix + [
+    hw_basename, hw_c, hw_o = _hw_filenames(task)
+    sds_cmd_head = prefix + [
         'sds++',
-        '-sds-pf', platform,
-        '-sds-hw', func_hw, c_hw, '-sds-end',
```
sampsyo: `func_hw` and `c_hw` are undefined here.

buildbot/buildbot/worker.py
```diff
@@ -312,13 +329,13 @@ def stage_hls(db, config):
...
         _task_config(task, config)
         flags = shlex.split(task['sdsflags'])
+        sds_cmd = _sds_cmd(prefix, task)
 
         # Run Xilinx SDSoC compiler for hardware functions.
         task.run(
-            _sds_cmd(prefix, hw_basename, hw_c, task['platform']) + flags + [
+            sds_cmd + flags + [
                 '-c',
                 hw_c, '-o', hw_o,
```
sampsyo: `hw_c` and `hw_o` are now undefined here.

buildbot/buildbot/worker.py
```diff
@@ -337,7 +354,7 @@ def stage_hls(db, config):
 
         # Run Xilinx SDSoC compiler for created objects.
         task.run(
-            _sds_cmd(prefix, hw_basename, hw_c, task['platform']) + flags + [
+            sds_cmd + flags + [
                 hw_o, HOST_O, '-o', config['EXECUTABLE_NAME'],
```
sampsyo: `hw_o` is also undefined here.

buildbot/buildbot/worker.py
```diff
@@ -285,17 +291,28 @@ def stage_seashell(db, config):
...
+        '-sds-pf', task['platform'],
+    ]
+    sds_cmd_tail = [
         '-clkid', '3',
         '-poll-mode', '1',
         '-verbose', '-Wall', '-O3',
     ]
+    if task['config']['directives']:
+        return sds_cmd_head + [
+            '-sds-hw', func_hw, c_hw, '-hls-tcl',
```
sampsyo: `func_hw` is undefined here.

buildbot/buildbot/worker.py
```diff
@@ -285,17 +291,28 @@ def stage_seashell(db, config):
...
+            task['config']['directives'], '-sds-end',
+        ] + sds_cmd_tail
+    else:
+        return sds_cmd_head + [
+            '-sds-hw', func_hw, c_hw, '-sds-end',
```
sampsyo: Also here.

sa2257:
> The non-Makefile path has some undefined variable references with this change (which would cause it to crash). Is it worth maintaining? We could just delete it and require Makefiles…

:scream: Thanks!

I'll test the normal flow and update the status.

sa2257:
For no directives, I still run with `directives=0`. A nice addition would be `-F directives=` to be different than no directive flag, as suggested above.

sa2257:
> Seems like this is ready to merge, if it works? If the `DIRECTIVES=0` hack works for some reason (i.e., the SDSoC compiler ignores directives files that don't exist, so the directive filename "0" suffices to mean "no directives" just as well as `DIRECTIVES=nothing_to_see_here` would), then I think we have a working solution.

That describes the current status well. I double checked whether `0` is actually meaningful to `sdsoc`, and it is not. So I'll merge this in.

Issue #211, opened 2019-07-08
Tracker: Benchmarking codegen hacks


sampsyo:
This issue is about things to change in code generation which are **just hacks** to make benchmarking easier. The idea is to avoid manual modifications to the output C++ code so we can directly evaluate what we generate. The spirit behind these hacks is *not* elegance; we just need to get the job done by any means necessary in the easiest possible way.

Here's what I am aware of so far:

- [x] *Software integers.* It's useful for debugging purposes to compile the "HLS" output code using a normal software compiler (i.e., gcc or clang). This means that the code needs to avoid using `ap_int` when doing that. I can see several ways to do that:
    - Insert an `#ifdef` (which activates only in an HLS compiler) around a `template` hack to replace `ap_int` with `int`.
    - Add a new backend mode, called like `vivado-sw` or something, that just generates `int` in place of `ap_int`.
    - Like the above, but use [`ac_int`](https://hlslibs.org/#collapseACDatatypes), as mentioned in https://github.com/cucapra/seashell/issues/105#issuecomment-475416029. (This would be nice but not in the spirit of a dumb-as-possible hack.)
- [x] *SDSoC pragmas.* We need to add the `#pragma SDS` nonsense that makes the whole SDSoC compiler toolchain go. An easy, inelegant way to do this would be to add a construct to the language that just specifies a string to output before the function declaration; we pass this string through unchanged.
- [x] *Trip count pragmas for estimation.* In estimation mode (but not when generating real hardware), the HLS tool wants to know the estimated trip count for `while` loops (which it obviously can't deduce by itself) so it can estimate the time the execution will take. We could also use the pass-through string approach for this.

Less important, but just nice-to-have features include:

- [ ] *Character literals.* See https://github.com/cucapra/fuse-benchmarks/pull/70#issuecomment-510670942. These should have type `bit<8>`.

sa2257:
For *Software integers* I'm adding 
```
#ifdef __SDSCC__
#include "ap_int.h"
#else
template < int N >
using ap_int = int;
#endif
```

sampsyo:
Moved character literals to #105. I'm closing this issue, since we've done the big ones—there are still outstanding differences in the manually-hacked C++, but we can open individual issues for those after we understand them.

PR #214, opened 2019-07-12
Evaluation hack: decorations for inserting #pragmas into generated source

sampsyo:
OK. I feel the need to apologize for how hacky this solution is before I even start explaining it.

It's a huge hack and I'm sorry.

There we have it.

---

As per #211 (items 2 and 3), this adds a new `decor` (i.e., decoration) command, whose purpose is to insert arbitrary strings that are then inserted in to the generated source code. They have no semantics. You do this:

```
decor "#pragma SDS data copy(m1[0:row_size][0:col_size])"
```

and it inserts that `#pragma` directly into the generated C++. The idea is that, for evaluation purposes, we want to insert some pragmas into the HLS source code that we *never* intend to support as first-class language features. They're tool-specific garbage that is not interesting from a language/type system perspective; they're just annotations necessary to make the complete evaluation work. This includes the SDSoC pragmas, IMO, because they're just about host/FPGA data movement—not at all about the actual logic HLS, which is what our language (currently) targets.

These `decor` constructs can appear in two locations:

- In the normal flow of commands (they're treated as a "block" command), in which case they get inserted (on their own line) at that point in the function. This is useful for loop annotations, for example, where you want to give a loop trip count for estimation purposes.
- At the top of the file, before the `decl`s, in which case they get emitted right before the C++ kernel function. This is useful for function-level pragmas, like the SDSoC one in my example above, that annotate the function arguments.

I also considered making these a little more complicated, so they'd be backend-sensitive—that is, you'd says something like `decl vivado "#pragma ..."` to insert an annotation *only* when generating Vivado code. But I thought that was a little too much too soon.

---

Here's an example. In our `machsuite-gemm-blocked` benchmark, I changed the file to look like this:

```
decor "#pragma SDS data copy(m1[0:row_size][0:col_size])"
decor "#pragma SDS data zero_copy(prod[0:row_size][0:col_size])"

decl m1: float[4096];
decl m2: float[4096];
decl prod: float[4096];

let temp_x:float = 0.0;
let mul:float = 0.0;

let i_row:bit<32> = 0;
let k_row:bit<32> = 0;

let jj:bit<32> = 0;
while (jj < 64) {
  let kk:bit<32> = 0;
  while (kk < 64) {
    for (let i = 0..64) {
      for (let k = 0..8) {
        i_row := i * 64;
        k_row := (k + kk) * 64;
        temp_x := m1[i_row + k + kk];
        for (let j = 0..8) {
          mul := temp_x * m2[k_row + j + jj];
          let temp = prod[i_row + j + jj];
          ---
          prod[i_row + j + jj] := temp + mul;
        }
      }
    }
    kk := kk + 8;
  }
  jj := jj + 8;
}
```

With this PR, that source now compiles to this:

```c++
// Avoid using `ap_int` in "software" compilation.
#ifdef __SDSVHLS__
#include "ap_int.h"
#else
template <int N> using ap_int = int;
#endif


#pragma SDS data copy(m1[0:row_size][0:col_size])
#pragma SDS data zero_copy(prod[0:row_size][0:col_size])
void gemm(float m1[4096], float m2[4096], float prod[4096]) {
  
  float temp_x = 0.0;
  float mul = 0.0;
  ap_int<32> i_row = 0;
  ap_int<32> k_row = 0;
  ap_int<32> jj = 0;
  while((jj < 64)) {
    ap_int<32> kk = 0;
    while((kk < 64)) {
      for(int i = 0; i < 64; i++) {
        for(int k = 0; k < 8; k++) {
          i_row = (i * 64);
          k_row = ((k + kk) * 64);
          temp_x = m1[(i_row + (k + kk))];
          for(int j = 0; j < 8; j++) {
            mul = (temp_x * m2[(k_row + (j + jj))]);
            float temp = prod[(i_row + (j + jj))];
            //---
            prod[(i_row + (j + jj))] = (temp + mul);
          }
        }
      }
      kk = (kk + 8);
    }
    jj = (jj + 8);
  }
}
```

That's pretty much what we want!

### DIFF
--- a/src/main/scala/common/Syntax.scala
+++ b/src/main/scala/common/Syntax.scala
@@ -193,6 +193,7 @@ object Syntax {
   case class CIf(cond: Expr, cons: Command, alt: Command) extends Command
   case class CFor(range: CRange, par: Command, combine: Command) extends Command
   case class CWhile(cond: Expr, body: Command) extends Command
+  case class CDecorate(value: String) extends Command
   case class CUpdate(lhs: Expr, rhs: Expr) extends Command {
     if (lhs.isLVal == false) throw UnexpectedLVal(lhs, "assignment")
   }
--- a/src/main/scala/common/Syntax.scala
+++ b/src/main/scala/common/Syntax.scala
@@ -225,7 +225,7 @@ object Syntax {
   case class Prog(
     includes: List[Include],
     defs: List[Definition],
-    decors: List[String],
+    decors: List[CDecorate],
     decls: List[Decl],
     cmd: Command) extends Positional
 
--- a/src/main/scala/common/Syntax.scala
+++ b/src/main/scala/common/Syntax.scala
@@ -225,6 +225,7 @@ object Syntax {
   case class Prog(
     includes: List[Include],
     defs: List[Definition],
+    decors: List[String],
     decls: List[Decl],
     cmd: Command) extends Positional
 
--- a/src/main/scala/passes/Checker.scala
+++ b/src/main/scala/passes/Checker.scala
@@ -59,7 +59,7 @@ object Checker {
     def checkLVal(e: Expr)(implicit env: Env): Env = checkE(e)
 
     def checkC(cmd: Command)(implicit env: Env): Env = cmd match {
-      case _:CSplit | _:CView | CEmpty => env
+      case _:CSplit | _:CView | CEmpty | CDecorate(_) => env
       case CPar(c1, c2) => checkCSeq(Vector(c1, c2))
       case CSeq(c1, c2) => checkCSeq(Vector(c1, c2))
       case CUpdate(lhs, rhs) => checkE(rhs)(checkLVal(lhs))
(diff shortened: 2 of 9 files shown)


rachitnigam:
It's a huge hack, but this totally seems the right way to do it! It also makes it easy to audit the uses of `decor` and see if we're using it as an escape hatch for something that should actually exist in the language.

src/main/scala/common/Syntax.scala
```diff
@@ -224,6 +225,7 @@ object Syntax {
   case class Prog(
     includes: List[Include],
     defs: List[Definition],
+    decors: List[String],
```
rachitnigam: Can this be `List[CDecorate]` to be consistent?

src/main/scala/passes/Checker.scala
```diff
@@ -59,7 +59,7 @@ object Checker {
     def checkLVal(e: Expr)(implicit env: Env): Env = checkE(e)
 
     def checkC(cmd: Command)(implicit env: Env): Env = cmd match {
-      case _:CSplit | _:CView | CEmpty => env
+      case _:CSplit | _:CView | CEmpty | CDecorate(_) => env
```
rachitnigam: Prefer `_: CDecorate` over exact matches of parameter lists.

src/main/scala/common/Syntax.scala
sampsyo: You're right; that's definitely better.

src/main/scala/passes/Checker.scala
sampsyo: Yep! Thanks.

PR #215, opened 2019-07-15
Support for unsigned bit numbers

rachitnigam:
- Signed and unsigned numbers are not comparable
- By default, signed numbers are inferred.

### DIFF
--- a/src/main/scala/typechecker/Subtyping.scala
+++ b/src/main/scala/typechecker/Subtyping.scala
@@ -53,7 +53,7 @@ object Subtyping {
   }
 
   def isSubtype(sub: Type, sup: Type): Boolean = (sub, sup) match {
-    case (TSizedInt(v1), TSizedInt(v2)) => v1 <= v2
+    case (TSizedInt(v1, un1), TSizedInt(v2, un2)) => un1 == un2 && v1 <= v2
     case (_:IntType, _:TSizedInt) => true
     case (_:TStaticInt, _:TIndex) => true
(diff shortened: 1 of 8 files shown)


src/test/scala/TypeCheckerSpec.scala
```diff
@@ -1278,6 +1278,16 @@ class TypeCheckerSpec extends FunSpec {
         let z: double = x + y;
         """ )
     }
+
+    it("unsigned and signed cannot be compared") {
```
sampsyo:
```suggestion
    it("unsigned and signed types are incomparable") {
```

(using the magic word "incomparable," which literally means the same thing but evokes partial ordering)

src/main/scala/typechecker/Subtyping.scala
```diff
@@ -53,7 +53,7 @@ object Subtyping {
   }
 
   def isSubtype(sub: Type, sup: Type): Boolean = (sub, sup) match {
-    case (TSizedInt(v1), TSizedInt(v2)) => v1 <= v2
+    case (TSizedInt(v1, un1), TSizedInt(v2, un2)) => un1 == un2 && v1 <= v2
```
sampsyo: Just for fun, I did a quick survey to see what other languages allow for conversions like this. C allows whatever, it seems, and both Go and Rust disallow both directions (i.e., they're incomparable).

sampsyo: Well-timed Regehr tweet: https://twitter.com/johnregehr/status/1150771871748509696

PR #216, opened 2019-07-15
Tiny fix: add type alias for `ap_uint` for SW compilation

sampsyo:
Just a little tweak in the spirit of #211. Following new support for `ubit` in #215, this extends our typedef-like hack that allows non-HLS compilation by mapping these to `unsigned int`.


rachitnigam:
Quick question: Can this header be turned into a `decor` that's included in every benchmark fuse file?

One way to do this:
- Define common.fuse.h where the only statement is a `decor` with the above pragma
- `import` the `common.fuse` in every fuse benchmark and compile it the header flag to elide the kernel.

sampsyo:
That’s a good idea. I’ll look into it.

Issue #218, opened 2019-07-15
Experiment: C level array parititoning vs partition pragma


rachitnigam:
## Experiment

**Hypothesis**: Manually partitioned arrays at C level perform as well as the partition pragma

**Program 1**: Two dimensional array with partition pragmas on both dimensions (say with 2 banks). Access each element of the array and increment it by 1 for 1000 iterations.

**Program 2**: 4 two dimensional arrays representing the partitions for the array in prog 1. A function with manual arbitration logic for selecting the right array based on the input arrays. Only use the function to access and update the elements of the arrays in the same way that prog 1 does.

## Rationale
I'm writing the Views section with a "Challenges in implementation" subsection which shows why the primitives exposed by various HLS backends are not the right one. 

One challenge in implementing views is we can't take slices of multidimensional arrays in C. We assume that this will be easy to do in a hypothetical RTL backend. Let's try to implement this at the C level and see how well it does and make sure we have an airtight argument backed up by an experiment.

Running this experiment also makes it easier to argue why views should be first class feature supported by the compiler.

@sampsyo can I ping you to do this? I'll write up the C code tonight if someone can try to push it through synthesis / estimation tomorrow.

sampsyo:
Good call! I would be interested to see this too. I really have no idea whether the tool’s built-in banking will have “magical” efficiency we can’t attain otherwise. Seems like we could even test this with a single-dimensional array, although a multi-dimensional one is of course where it really matters for our backends.

Happy to run some jobs when I get a moment. But FWIW, here is my favorite way to run jobs from home:

    $ ssh -NL8000:gorgonzola:8000 waffle
    $ export BUILDBOT=http://localhost:8000

rachitnigam:
Tracked in https://github.com/cucapra/fuse-benchmarks/pull/93

PR #219, opened 2019-07-16
Add partition experiment

rachitnigam:
C++ files for experiment described in #218.

Works with `make SOFTWARE=1` in both directories.


sampsyo:
Added in this branch, with more fixes to make hardware execution actually work:
https://github.com/cucapra/fuse-benchmarks/tree/experiment-partitioning

sampsyo:
Status so far:
- [manual-partition](http://localhost:8000/jobs/lMlAqThrvQ8.html) fails in estimation because the loop bounds need estimated trip count annotations.
- [using-pragma](http://localhost:8000/jobs/oTYgB2DPJYM.html) fails in HLS compilation because of more damn errors in the generated SDSoC host-side interface code. :cry: :scream: :angry:

(Those are links to proxied Buildbot pages, using my SSH tunnel trick described elsewhere.) This experiment can still work, but it will take some hammering to get it to go through SDSoC estimation.

rachitnigam:
Okay, feel free to close this and discuss this in fuse-benchmarks instead.

> manual-partition fails in estimation because the loop bounds need estimated trip count annotations.

Is this with the top `while` loop? We can make `iterations` be a constant to make it work.

> using-pragma fails in HLS compilation because of more damn errors in the generated SDSoC host-side interface code.

Should I try porting to SDAccel?

sampsyo:
OK! No need to move the discussion, but I did merge the (tweaked) code into `master` for convenience.

> Is this with the top `while` loop? We can make `iterations` be a constant to make it work.

Yes—although probably the right thing to do is to add estimated trip-count pragmas, like this:
https://github.com/cucapra/fuse-benchmarks/blob/ee826630c072347a7371c8df92a63db23a6eec81/rewrite/machsuite-viterbi/viterbi.cpp#L23

That is, something like this:

```
while (...) {
    decor "#pragma HLS loop_tripcount avg=42"
    ....
```

> Should I try porting to SDAccel?

Good question, but my intuition says no—I get the sense this should be debuggable, with a little fiddling with the function signature…

sampsyo:
Also for future convenience, here's a copy-n-pastable command to check out tweaks:

```
./_scripts/batch.py experiments/partition/manual-partition experiments/partition/using-pragma
```

rachitnigam:
Tracked in the benchmarks repo

PR #228, opened 2019-07-29
buildbot toolchain configuration

ViviYe:
edited buildbot to work with both f1 instance and sdsoc.

### DIFF
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -61,21 +61,16 @@ LOG_PREVIEW_LINES = 32
 COMPILE_TIMEOUT = 120
 SYNTHESIS_TIMEOUT = 9000
 
-# Default Xilinx target platform.
-DEFAULT_PLATFORM = 'zed'
-# Default F1 target 
-DEFAULT_F1_TARGET = 'sw_emu'
-
-
-# Global switch for configuring buildbot. 
-# If set to f1, the buildbot is configured to run on f1. 
-# Otherwise, buildbot will use the sdsoc workflow.
+# The Buildbot currently supports two backend toolchains: Xilinx's SDSoC
+# (for Zynq processors) and SDAccel (for AWS F1). Set this to "f1" for
+# deployment on F1; leave it as anything else for the SDSoC workflow.
 TOOLCHAIN = 'f1'
 
-# The S3 bucket to be used for f1 instance 
-# bucket-name
+# Options for SDSoC/Zynq.
+DEFAULT_PLATFORM = 'zed'  # Use the "platform" job config option to override.
+
+# Options for SDAccel/F1.
+DEFAULT_F1_TARGET = 'sw_emu'  # Use the "target" job config option to override.
 S3_BUCKET = 'test-bucket-1025132741'
-# dcp-folder-name
-S3_DCP = 'DCPs'
-# logs-folder-name
-S3_LOG = 'SDAccel_log'
+S3_DCP = 'DCPs'  # dcp-folder-name
+S3_LOG = 'SDAccel_log'  # logs-folder-name
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -74,6 +74,4 @@ DEFAULT_F1_MODE = 'sw_emu'  # Use the "mode" job config option to override.
 S3_BUCKET = 'test-bucket-1025132741'
 S3_DCP = 'DCPs'  # dcp-folder-name
 S3_LOG = 'SDAccel_log'  # logs-folder-name
-
-#sleep time between each AFI status check
-AFI_CHECK_INTERVAL = 300 
+AFI_CHECK_INTERVAL = 300  # Sleep time between each AFI status check.
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -74,3 +74,6 @@ DEFAULT_F1_MODE = 'sw_emu'  # Use the "mode" job config option to override.
 S3_BUCKET = 'test-bucket-1025132741'
 S3_DCP = 'DCPs'  # dcp-folder-name
 S3_LOG = 'SDAccel_log'  # logs-folder-name
+
+#sleep time between each AFI status check
+AFI_CHECK_INTERVAL = 300 
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -46,7 +46,7 @@ CONFIG_OPTIONS = {
     'hwname': str,
     'sdsflags': str,
     'platform': str,
-    'target': str,
+    'mode': str,
 }
 
 # The name to use for compiled executables.
@@ -70,7 +70,7 @@ TOOLCHAIN = 'f1'
 DEFAULT_PLATFORM = 'zed'  # Use the "platform" job config option to override.
 
 # Options for SDAccel/F1.
-DEFAULT_F1_TARGET = 'sw_emu'  # Use the "target" job config option to override.
+DEFAULT_F1_MODE = 'sw_emu'  # Use the "mode" job config option to override.
 S3_BUCKET = 'test-bucket-1025132741'
 S3_DCP = 'DCPs'  # dcp-folder-name
 S3_LOG = 'SDAccel_log'  # logs-folder-name
--- a/buildbot/buildbot/config_default.py
+++ b/buildbot/buildbot/config_default.py
@@ -15,6 +15,10 @@ HLS_COMMAND_PREFIX = []
 # development environment and False in production.
 WORKER_THREADS = None
 
+# The number of jobs to process in parallel in the "make" stage (which is the
+# expensive, long-running one).
+PARALLELISM_MAKE = 1
+
 # Filename extensions to send as plain text for job file viewing.
 TEXT_EXTENSIONS = [
     'ss',
@@ -42,6 +46,7 @@ CONFIG_OPTIONS = {
     'hwname': str,
     'sdsflags': str,
     'platform': str,
+    'target': str,
 }
 
 # The name to use for compiled executables.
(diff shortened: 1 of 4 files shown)


buildbot/buildbot/config_default.py
```diff
@@ -54,7 +58,13 @@
...
+SYNTHESIS_TIMEOUT = 9000
 
 # Default Xilinx target platform.
 DEFAULT_PLATFORM = 'zed'
+DEFAULT_TARGET = 'sw_emu'
+TOOLCHAIN = 'f1'
+
+S3_BUCKET = 'test-bucket-1025132741'
+S3_DCP = 'DCPs'
+S3_LOG = 'SDAccel_log'
```
sampsyo: Some comments here might be nice, to describe what these options do! For example, `TOOLCHAIN` is a big, important switch that people should be aware of (and which options are available). And `DEFAULT_TARGET` only matters for F1 (i.e., SDAccel), right? Same with the S3 options?

buildbot/buildbot/server.py
```diff
@@ -33,6 +33,8 @@
     state.HLS_FINISH: "Synthesized",
     state.MAKE: "make",
     state.MAKE_PROGRESS: "make-ing",
+    state.AFI_START:"starting AFI generation",
+    state.AFI:"generating AFI",
```
sampsyo: Following the pattern of the other names here, maybe these could be called "Starting AFI" and "Generating AFI." (Also, this is a tiny thing, but it's customary to put a space after the `:`. :smiley:)

buildbot/buildbot/state.py
```diff
@@ -34,3 +34,5 @@
 RUN = "fpga_executing"
 DONE = "done"
 FAIL = "failed"
+AFI_START = "starting AFI generation"
+AFI = "generating AFI"
```
sampsyo: Perhaps add these to the ASCII art diagram above? 🖌

buildbot/buildbot/worker.py
```diff
@@ -214,20 +223,35 @@ def stage_make(db, config):
...
         _task_config(task, config)
 
         if task['sdsflags']:
             task.log('WARNING: make stage is ignoring sdsflags={}'.format(
                 task['sdsflags']
             ))
-
-        make_cmd = prefix + [
+        # if toochain set to f1, use different make command.
+        if (TOOLCHAIN == 'f1'):
```
sampsyo: Instead of importing the `TOOLCHAIN` variable, use `config['TOOLCHAIN']`. (Otherwise, you'll always get the default value, even if someone overrides it in their local config.)

No parentheses are necessary for `if`s in Python.

buildbot/buildbot/worker.py
```diff
@@ -70,12 +73,10 @@ def set_state(self, state):
 
     def run(self, cmd, capture=False, timeout=60, cwd='', **kwargs):
         """Run a command and log its output.
-
         Return an exited process object. If `capture`, then  the
         standard output is *not* logged and is instead available as the return
         value's `stdout` field. Additional arguments are forwarded to
         `subprocess.run`.
-
```
sampsyo: Did these blank lines get deleted accidentally?

buildbot/buildbot/worker.py
```diff
@@ -375,10 +399,66 @@ def stage_hls(db, config):
...
+    """AFI stage: Create the AWS FPGA binary and AFI from the *.xclbin 
+    (Xilinx FPGA binary file)
+    """
+    with work(db, state.AFI_START, state.AFI, state.HLS_FINISH) as task:
+        # clean all the generated files from previous runs.
+        task.run(
+            ['sh', '-c', 'cd xclbin; \
+              rm -rf to_aws; rm -rf *afi_id.txt;\
+              rm -rf *.tar; rm -rf *afgi_id.txt'],
+            cwd=CODE_DIR
```
sampsyo: Instead of using `cd xclbin ; ...` you can use `cwd=os.path.joing(CODE_DIR, 'xclbin')` here.

buildbot/buildbot/worker.py
```diff
@@ -375,10 +399,66 @@ def stage_hls(db, config):
...
+def stage_afi(db, config):
+    """AFI stage: Create the AWS FPGA binary and AFI from the *.xclbin 
+    (Xilinx FPGA binary file)
+    """
+    with work(db, state.AFI_START, state.AFI, state.HLS_FINISH) as task:
+        # clean all the generated files from previous runs.
+        task.run(
+            ['sh', '-c', 'cd xclbin; \
+              rm -rf to_aws; rm -rf *afi_id.txt;\
+              rm -rf *.tar; rm -rf *afgi_id.txt'],
```
sampsyo: Instead of using multiple `rm` commands, you can put them all in one command, like `rm -rf to_aws *afi_id.txt ...`.

buildbot/buildbot/worker.py
```diff
@@ -375,10 +399,66 @@ def stage_hls(db, config):
...
+            cwd=CODE_DIR
+        )
+        # find the hw synthesis xclbin file.
+        xclbin_file_path = task.run(
+            ['sh', '-c', 'find ./xclbin  -type f -iname "*hw\.*.xclbin"'],
```
sampsyo: This probably doesn't need a shell wrapper. You can just do `['find', './xclbin', '-type', '-f', ...]`. Or, it may even be possible to find the file through "normal means" without even executing a shell by using Python's [glob](https://docs.python.org/3/library/glob.html) module.

buildbot/buildbot/worker.py
```diff
@@ -375,10 +399,66 @@ def stage_hls(db, config):
...
+        )
+        # loops every 5 minutes to check if the AFI is ready.
+        while True:
+            time.sleep(300)
+            afi_id_file = task.run(
+                ['sh', '-c', 'cat ./xclbin/*afi_id.txt'],
+                cwd=CODE_DIR,
+                capture=True
+            )
+            afi_json = json.loads(afi_id_file.stdout.decode('utf8').strip())
```
sampsyo: Instead of using a shell to `cat` the file, you can read it directly, if you know the filename (which you might want to use glob for again):

```py
with open('./xclbin/foo_afi_id.txt') as f:
    afi_json = json.load(f)
```

buildbot/buildbot/worker.py
```diff
@@ -375,10 +399,66 @@ def stage_hls(db, config):
...
+            afi_id = afi_json['FpgaImageId']
+            status_string = task.run(
+                ['sh', '-c',
+                    'aws ec2 describe-fpga-images \
+                     --fpga-image-ids {}'.format(afi_id)],
```
sampsyo: No shell or formatting necessary here: `['aws', 'ec2', 'describe-fpga-images', '--fpga-image-ids', afi_id]` should work.

buildbot/buildbot/worker.py
```diff
@@ -375,10 +399,66 @@ def stage_hls(db, config):
...
+                cwd=CODE_DIR,
+                capture=True
+            )
+            status_json = json.loads(
+                status_string.stdout.decode('utf8').strip())
```
sampsyo: I think the `json` module can load bytestrings, so the `decode` and `strip` steps here should probably be unnecessary.

buildbot/buildbot/worker.py
```diff
@@ -375,10 +399,66 @@ def stage_hls(db, config):
...
+            # when the afi becomes available, exit the loop and enter
+            # execution stage.
+            if status_json['FpgaImages'][0]['State']['Code'] == 'available':
```
sampsyo: It might be nice to include a `task.log` here to record what the status is?

buildbot/buildbot/worker.py
```diff
@@ -423,6 +516,6 @@ def work_threads(db, config):
     """
     out = []
     for stage in (stage_unpack, stage_make, stage_seashell,
-                  stage_hls, stage_fpga_execute):
+                  stage_hls, stage_afi, stage_fpga_execute):
```
sampsyo: Maybe we should *only* include `stage_afi` when `config['TOOLCHAIN'] == 'f1'`.

sampsyo:
Nice! I have made a few low-level improvements here and there. (I also rebased on master.) @ViviYe, can you please take a look and see if I messed anything up?

Here's a short wish list of little improvements too, but they're obviously not high priorities:

- The `TARGET` Makefile variable is sort of overloaded between the two Makefile paths (i.e., it means two different things). That's unfortunate—maybe we could pick a different name for one or the other?
- It would be great if the `estimate` config option worked on both toolchains…
- For the `task.run` invocations with `sh -c`, it's probably possible to pass `shell=True` instead and just provide the shell string for marginally more readable code.
- We could make the "sleep time" for checking the AFI status a config option.

ViviYe:
Thank you so much!!!! It looks great! I can definitely work on the list of small improvements.

ViviYe:
I am also not sure what estimate option should do on f1-instance?

sampsyo:
Yeah, I'm not sure either. Do I recall correctly that there is no command-line flag for producing performance estimates? If so, I guess we're out of luck. But maybe we'll find one someday…

ViviYe:
I don't think I saw any command line that does performance estimate but I can look into it? I am also not sure: for changing the 'TARGET' variable to something else, should I only change it in the benchmarks or also in the buildbot? Would it be a bit confusing to have make TARGET = something else in the make stage?

sampsyo:
That’s a good question... I don’t think it’s super important either, but I think the most reasonable thing would be to change the name of TARGET in the Makefile for F1, and then use that name in the Buildbot.

ViviYe:
I just changed TARGET variable to MODE(I hope it make sense), added sleep time configuration and changed all the sh -c call to shell = True! Do you mind taking a look and see if all the changes i made make sense when you have time? Thank you so much! !! 😆

sampsyo:
Looks perfect! I pushed just a few minor code/comment style improvements. :tada: :confetti_ball:

I think this is basically ready to merge. When we do, we should be sure to configure the local Buildbot to _not_ use the F1 toolchain. I'll do that preemptively!

ViviYe:
YAY! Thank you so much!!

PR #229, opened 2019-07-31
Simplify the C++ preamble for Vivado backend

sampsyo:
Given the hack in https://github.com/cucapra/fuse-benchmarks/pull/109, we will have a drop-in replacement for this header in software, so we won't need the #ifdef in generated code.


rachitnigam:
LGTM

PR #233, opened 2019-08-05
Toolchain config

ViviYe:
Fixed the bug in execution stage.
increased the timeout for hw_emu execution and deleted shell=true for hw execution


sampsyo:
Yo; thanks for doing this! However, to get something that can be cleanly merged, you'll want to base your new commits on master instead of your old branch. I think the easiest way to do this now would be to create a _new_ branch off of master and use `git cherry-pick` on the _new_ commits you want to move them to the new branch. Then open a separate PR from that new branch.

rachitnigam:
@ViviYe is this branch stale or are you planning on merging changes from this at some point?

sampsyo:
Superseded by #228!

PR #237, opened 2019-08-15
Worst-effort pipelining

sampsyo:
This Fuse code:

```
for (let i = 0..4) pipeline {
  let a = 1 + 2;
  let b = 3 + 4;
}
```

works and generated HLS code with a `#pragma HLS PIPELINE` annotation. This version:

```
for (let i = 0..4) pipeline {
  let a = 1 + 2
  ---
  let b = 3 + 4
}
```

fails with a message saying that pipelining is only allowed on non-sequenced loops.

### DIFF
--- a/src/main/scala/Parser.scala
+++ b/src/main/scala/Parser.scala
@@ -216,7 +216,7 @@ private class FuseParser extends RegexParsers with PackratParsers {
     block |
     cfor |
     conditional |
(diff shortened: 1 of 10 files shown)


rachitnigam:
[Worse is better](https://web.mit.edu/6.033/www/papers/Worse_is_Better.pdf).

PR #240, opened 2019-08-19
executable name can now be customized

ViviYe:
set a new variable f1-executable-name (currently set to sdaccel) instead of assuming the executable name is host and run ./host.

### DIFF
--- a/buildbot/buildbot/worker.py
+++ b/buildbot/buildbot/worker.py
@@ -487,7 +487,7 @@ def stage_fpga_execute(db, config):
             task.log('skipping FPGA execution stage')
             return
 
-       if config['TOOLCHAIN'] == 'f1':
+        if config['TOOLCHAIN'] == 'f1':
             # On F1, use the run either the real hardware-augmented
             # binary or the emulation executable.
             if task['mode'] == 'hw':
--- a/buildbot/buildbot/worker.py
+++ b/buildbot/buildbot/worker.py
@@ -493,13 +493,13 @@ def stage_fpga_execute(db, config):
             if task['mode'] == 'hw':
                 exe_cmd = ['sudo', 'sh', '-c',
                            'source /opt/xilinx/xrt/setup.sh ;\
-                            ./{}'.format(config['F1_EXECUTABLE_NAME'])]
+                            ./{}'.format(config['EXECUTABLE_NAME'])]
             else:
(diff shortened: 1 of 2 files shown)


sampsyo:
Seems cool! But maybe we should just have one `EXECUTABLE_NAME` option—there's no particular reason why the SDSoC executable is called `sdsoc`, for example. It could just be called `exe` by default. :smiley:

ViviYe:
Should I just change the executable name to exe in both builedbot and in fuse-benchmarks? 😀

sampsyo:
Hmm; I'm not sure what you mean by changing it in fuse-benchmarks… the way this works for SDSoC, at least, is that the configured `EXECUTABLE_NAME` here just overrides the default specified in the Makefile. So the Makefile can use whatever sensible name it wants! The Buildbot just overrides it with one that it knows how to invoke.

ViviYe:
Thanks so much! That makes sense! Just being curious, how do the variables set in buildbot override variables in Makefile?

sampsyo:
They work by invoking the `make` command with its special variable-setting argument syntax, like this:

```
make FOO=bar
```

That command invokes the `Makefile` in the current directory, while setting the variable `$FOO` to `bar`. Even if that file contains an assignment:

```
FOO := baz
```

Then the assignment will be ignored, and `$FOO` will never contain `baz`—only `bar`.

ViviYe:
Thanks so much!!!

ViviYe:
Just changed sdaccel also to use the same ```EXECUTABLE_NAME``` variable!

buildbot/buildbot/worker.py
```diff
@@ -492,15 +492,20 @@ def stage_fpga_execute(db, config):
             # binary or the emulation executable.
             if task['mode'] == 'hw':
                 exe_cmd = ['sudo', 'sh', '-c',
-                           'source /opt/xilinx/xrt/setup.sh ; ./host']
+                           'source /opt/xilinx/xrt/setup.sh ;\
+                            ./{}'.format(config['EXECUTABLE_NAME'])]
             else:
-                exe_cmd = ['sh', '-c', 'cur=`pwd`; cd $AWS_FPGA_REPO_DIR ;\
+                exe_cmd = ['sh','-c','cur=`pwd`; cd $AWS_FPGA_REPO_DIR ;\
```
sampsyo: Were these spaces deleted by mistake? :smiley:

ViviYe: sorry I didn't notice this! Just fixed it! 😭

PR #242, opened 2019-08-22
Minor cleanup for F1 bb code

rachitnigam:
Some minor fixes for f1 code.
- `source` can directly refer to the file path. No need to `cd` into a directory unless something weird is going on with F1. 

- `afi` file is generated right after the AFI stage starts (@ViviYe correct?). Just read the value once when polling for the command to finish.

### DIFF
--- a/buildbot/buildbot/worker.py
+++ b/buildbot/buildbot/worker.py
@@ -235,15 +235,13 @@ def stage_make(db, config):
             ))
 
         if config['TOOLCHAIN'] == 'f1':
-            # Get the AWS platform ID for F1 builds.
-            platform_script = (
-                'cd $AWS_FPGA_REPO_DIR ; '
-                'source ./sdaccel_setup.sh > /dev/null ; '
-                'echo $AWS_PLATFORM'
+            # Configure current environment.
+            source_script = (
+                'source $AWS_FPGA_REPO_DIR/sdaccel_setup.sh ; '
             )
-            proc = task.run([platform_script], capture=True, shell=True)
-            aws_platform = proc.stdout.decode('utf8').strip()
+            proc = task.run([source_script], capture=True, shell=True)
 
+            aws_platform = os.environ('AWS_PLATFORM')
             make = [
                 'make',
                 'MODE={}'.format(task['mode']),
--- a/buildbot/buildbot/worker.py
+++ b/buildbot/buildbot/worker.py
@@ -492,9 +492,8 @@ def stage_fpga_execute(db, config):
                 exe_cmd = ['sudo', 'sh', '-c',
                            'source /opt/xilinx/xrt/setup.sh ; ./host']
             else:
-                exe_cmd = ['sh', '-c', 'cur=`pwd`; cd $AWS_FPGA_REPO_DIR ;\
-                source ./sdaccel_setup.sh > /dev/null; \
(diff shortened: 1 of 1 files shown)


buildbot/buildbot/worker.py
```diff
@@ -235,15 +235,13 @@ def stage_after_make(task):
...
-                'echo $AWS_PLATFORM'
+            # Configure current environment.
+            source_script = (
+                'source $AWS_FPGA_REPO_DIR/sdaccel_setup.sh ; '
             )
-            proc = task.run([platform_script], capture=True, shell=True)
-            aws_platform = proc.stdout.decode('utf8').strip()
+            proc = task.run([source_script], capture=True, shell=True)
 
+            aws_platform = os.environ('AWS_PLATFORM')
```
sampsyo: Sadly, I don't this will quite work. `os.environ` is the environment of the _current_ process, i.e., the buildbot worker. We need the variable from the _subprocess_, i.e., the shell that just ran the `source` command. That's the reason for the in-shell `echo`.

buildbot/buildbot/worker.py
```diff
@@ -494,9 +492,9 @@ def stage_fpga_execute(db, config):
                 exe_cmd = ['sudo', 'sh', '-c',
                            'source /opt/xilinx/xrt/setup.sh ; ./host']
             else:
-                exe_cmd = ['sh', '-c', 'cur=`pwd`; cd $AWS_FPGA_REPO_DIR ;\
-                source ./sdaccel_setup.sh > /dev/null; \
-                cd $cur; XCL_EMULATION_MODE={} ./host'.format(task['mode'])]
+                exe_cmd = ['sh', '-c', 'source $AWS_FPGA_REPO_DIR/sdaccel_setup.sh > /dev/null;\
+                XCL_EMULATION_MODE={} ./host'.format(task['mode'])]
```
sampsyo: I think the reason we need the `cur=... ; ...; cd $cur` dance is that the setup script assumes it will be `source`'d from its containing directory. But perhaps I'm wrong—does this work without the `cd` dance?

sampsyo: Unrelatedly, it occurs to me that we could probably pass `XCL_EMULATION_MODE` via an `env` parameter to `task.run` as opposed to interpolating it into the shell line.

buildbot/buildbot/worker.py
rachitnigam: Ah, thanks for the explanation! Reverted.

ViviYe:
Yes the afi file is generated right after the AFI stage starts! The changes do make sense since we only need to get the afi_id once. Thanks so much for helping me clean up the code!

buildbot/buildbot/worker.py
ViviYe: I just did the experiment on f1 instance and it seems like the setup script does not need to be in the directory where it is sourced from! 😲

sampsyo: Excellent news!

rachitnigam: Amazing! We can merge it whenever @Vivi verifies this doesn’t break everything (doesn’t have to be right now!)

ViviYe:
The changes look good! I just tested this buildbot on f1 and it does not break anything!

PR #244, opened 2019-08-23
Rebase #240 on master

rachitnigam:
Clean rebase of #240 on master. Seems like the only change is the timeout. @ViviYe, if this looks correct, merge this branch and close #240.

If my comment doesn't make sense, feel free to ask before merging.

### DIFF
--- a/buildbot/buildbot/worker.py
+++ b/buildbot/buildbot/worker.py
@@ -500,7 +500,7 @@ def stage_fpga_execute(db, config):
             task.run(
                 exe_cmd,
                 cwd=CODE_DIR,
-                timeout=1800
+                timeout=3600
             )
 
         else:


sampsyo:
Hmm; I'm not sure what went wrong with the rebase, but that older PR does have other outstanding stuff—for example, this line on master should set the config variable to `'exe'` post-PR, rather than `'sdsoc'`:
https://github.com/cucapra/seashell/blob/ea903e383620210cdb07f705474ae6bd4cbbe58a/buildbot/buildbot/config_default.py#L53

rachitnigam:
Oh, odd. I see the discrepancy. I think the confused he line changes in #240 because I didn’t notice they were changing the host invocation to exe invocation.

PR #245, opened 2019-08-23
Cleanup buildbot for move

rachitnigam:
Addresses two comments from @sampsyo for BB move:

- Remove fuse specific stages from BB.
- Remove the silly live interface from BB.
- Address a few unused variable issues from pylint.

Not tested live.


buildbot/README.md
```diff
@@ -66,7 +64,8 @@ If the directory contains data files with `.data` extension, they'll be copied o
 
 You can also specify job configuration options as further POST parameters:
 
-- `skipseashell`, which lets you supply plain HLS C code as input.
+**NOTE(rachit)**: These options are out of date. Document them properly here.
```
sampsyo: I think they’re all right! We’re missing a few, but the ones that are here look fine to me.

buildbot/buildbot/workproc.py
```diff
@@ -40,7 +40,7 @@ def start(self):
             if not thread.is_alive():
                 thread.start()
 
-    async def handle(self, client, addr):
+    async def handle(self, client, _):
```
sampsyo: Not that it actually matters, but I’d leave the name of the unused parameter here in situations like this (and the config parameter above). Because these are both sort of callbacks from an existing API, I kind of like to know what information the API is providing to the function—even if the current version of the code doesn’t need it. It saves an extra trip to the docs to figure out what the unlabeled information is, and it makes it clear that the shape of the function matches all the other functions that play a similar role.

buildbot/buildbot/worker.py
```diff
@@ -532,8 +479,7 @@ def stage_fpga_execute(db, config):
             )
 
 
-STAGES = (stage_unpack, stage_make, stage_seashell, stage_hls,
-          stage_fpga_execute)
+STAGES = (stage_unpack, stage_make, stage_hls, stage_fpga_execute)
```
sampsyo: No need for parentheses if it all fits on one line.

buildbot/buildbot/worker.py
```diff
@@ -104,7 +103,7 @@ def run(self, cmd, capture=False, timeout=60, cwd='', **kwargs):
                     exc.filename,
                 ))
             except subprocess.TimeoutExpired as exc:
-                raise WorkError('timeout after {} seconds{}'.format(
+                raise WorkError('timeout after {} seconds'.format(
```
sampsyo: Weird! Not sure how that slipped in here.

buildbot/buildbot/worker.py
rachitnigam: Woah, didn't know you define arrays like that!

PR #246, opened 2019-08-23
Remove Kiama dependency

rachitnigam:
- Add Document class from deprecated scala.text into source.
- Will allow for 2.13 migration and usage of scala.js.


rachitnigam:
Also migrates the project to 2.13

PR #248, opened 2019-08-24
Move buildbot to separate repo

### LINKED ISSUE #32: Buildbot tracker

### LINKED ISSUE #210: Make the Buildbot independent
I don't know why this realization came to me only now, but I have come to a startling epiphany about the soul of the Buildbot.

I originally made it to address three problems:
- tools that are super slow and therefore annoying to run on your dev machine
- tools that are inconvenient to run at all because of licensing and platform dependencies
- the need for access to special hardware

None of these problems apply to the Seashell compiler itself. It's all about dealing with HLS and other FPGA tools. It's not inconvenient, slow, or expensive to run our compiler locally (despite Scala :trollface:). At this point, we're even compiling everything locally before submitting jobs anyway.

So I think we should separate the Buildbot from the compiler. In its soul, the Buildbot doesn't want to be Seashell-specific—it just wants to be a convenient wrapper for inconvenient FPGA tools and hardware. We can give it a nice name, like Cyclops or Polyphemus, and send it on its way. Freed from the need to compile Seashell programs, it can be simpler and better fulfill its true purpose—wrapping the FPGA toolchains.

Concretely, this means we should:
- move it to a separate git repository
- drop the Seashell stage (we can keep the HLS stage)
- remove the silly "live interface" and the "non-Seashell" config option
- simplify the Docker setup on our server

This will be satisfying for me to do, so I'll plan on doing it. It is, however, not a terribly high priority and we have bigger fish to fry at the moment.

rachitnigam:
Fixes #210.
Fixes #32.


sampsyo:
FYI, the README still mentions a `buildbot/` directory.

PR #256, opened 2019-09-23
Sized check

tissue3:
Check if static int is the right sized int type

### DIFF
--- a/src/main/scala/Parser.scala
+++ b/src/main/scala/Parser.scala
@@ -29,7 +29,8 @@ private class FuseParser extends RegexParsers with PackratParsers {
   lazy val uInt: P[Expr] = "(-)?[0-9]+".r ^^ { n => EInt(n.toInt) }
   lazy val hex = "0x[0-9a-fA-F]+".r ^^ { n => Integer.parseInt(n.substring(2), 16) }
   lazy val octal = "0[0-7]+".r ^^ { n => Integer.parseInt(n.substring(1), 8) }
-  lazy val float = "(-)?[0-9]+\\.[0-9]+".r
+  //lazy val fixed = "f'(-)?[0-9]+\\.[0-9]+".r 
+  lazy val double = "(-)?[0-9]+\\.[0-9]+".r ^^ {n => EDouble(n.toDouble)}
   lazy val boolean = "true" ^^ { _ => true } | "false" ^^ { _ => false }
 
   lazy val arrLiteral: P[Expr] = positioned {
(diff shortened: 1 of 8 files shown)


src/test/scala/TypeCheckerSpec.scala
```diff
@@ -413,13 +413,13 @@ class TypeCheckerSpec extends FunSpec {
       typeCheck("""
               let bucket_idx = 10;
               ---
-              bucket_idx := 20;
+              bucket_idx := (20 as bit<4>);
```
rachitnigam: You’d probably want to give a larger type to bucket_idx above.

src/main/scala/Parser.scala
```diff
@@ -29,7 +29,8 @@ private class FuseParser extends RegexParsers with PackratParsers {
   lazy val uInt: P[Expr] = "(-)?[0-9]+".r ^^ { n => EInt(n.toInt) }
   lazy val hex = "0x[0-9a-fA-F]+".r ^^ { n => Integer.parseInt(n.substring(2), 16) }
   lazy val octal = "0[0-7]+".r ^^ { n => Integer.parseInt(n.substring(1), 8) }
-  lazy val float = "(-)?[0-9]+\\.[0-9]+".r
+  //lazy val fixed = "f'(-)?[0-9]+\\.[0-9]+".r
```
rachitnigam: Doesn’t need to be in this PR.

src/main/scala/typechecker/Subtyping.scala
```diff
@@ -54,7 +55,9 @@ object Subtyping {
 
   def isSubtype(sub: Type, sup: Type): Boolean = (sub, sup) match {
     case (TSizedInt(v1, un1), TSizedInt(v2, un2)) => un1 == un2 && v1 <= v2
-    case (_:IntType, _:TSizedInt) => true
+    case (TStaticInt(v1), TSizedInt(v2,un2)) => ( (v1<0 && un2==false) | (v1>=0) ) && bitsNeeded(v1)<=v2 
+    case (_:TIndex,_:TSizedInt) => true   
+    //case (_:IntType, _:TSizedInt) => true
```
rachitnigam: You can just delete this.

src/test/scala/TypeCheckerSpec.scala
tissue3: Yes, that's not the usual usage of as cast, but I leave it here to remind reader 10 is implicitly viewed as 4 bit number. Another example of usual usage just follows it (let test_var:bit<32> = 10;).

src/main/scala/typechecker/Subtyping.scala
tissue3: Sure wait for next PR.

PR #257, opened 2019-09-25
Sized check

tissue3:
This should fix fixed point.


src/main/scala/common/Syntax.scala
```diff
@@ -70,8 +73,11 @@ object Syntax {
   // Use case class instead of case object to get unique positions
   case class TVoid() extends Type
   case class TBool() extends Type
+
+  case class TRational(value:Double) extends Type
```
sampsyo: This is a pretty fine-grained thing and probably doesn't matter too much, but we probably want to represent the underlying value precisely—i.e., in the same way that `ERational` represents the number, which is currently a string. (You could also imagine representing it as a pair of integers, the integer and fractional parts, but a string is basically just as good.)

src/main/scala/typechecker/TypeCheck.scala
```diff
@@ -470,6 +471,7 @@ object TypeChecker {
       // have static types.
       val rTyp = typ.map(env.resolveType(_) match {
         case TStaticInt(v) => TSizedInt(bitsNeeded(v), false)
+        case _:TRational => TDouble()
```
sampsyo: Here (and the next addition below in this diff) is where we're providing the default for rational literals. It is probably a good idea to extend the comment above to describe what we're doing, i.e., providing a default when we can't otherwise possibly know what the right numerical type is.

src/main/scala/typechecker/Subtyping.scala
```diff
@@ -57,32 +57,50 @@ object Subtyping {
...
+      case None => 
+        if (bitsNeeded(v1.toInt)>bitsNeeded(v2.toInt)) Some(TRational(v1.toInt)) 
+        else Some(TRational(v2.toInt)) 
+    }
     case (TSizedInt(s1, un1), TSizedInt(s2, un2)) =>
       if (un1 == un2) Some(TSizedInt(max(s1, s2), un1))
       else None
     case (TSizedInt(s, un), TStaticInt(v)) =>
-      Some(TSizedInt(max(s, bitsNeeded(v)), un))
+      Some(TSizedInt(max(s, bitsNeeded(v)), un))
```
sampsyo:
```suggestion
      Some(TSizedInt(max(s, bitsNeeded(v)), un))
```

src/main/scala/backends/CppRunnable.scala
```diff
@@ -23,7 +23,7 @@ private class CppRunnable extends CppLike {
     case _:TIndex | _:TStaticInt => text("int")
     case TSizedInt(_, un) => text(if (un) "unsigned int" else "int")
     case _:TFloat => text("float")
-    case _:TDouble => text("double")
+    case _:TDouble | _:TRational | _:TFixed => text("double")
```
sampsyo: Codegen should probably never need to give a type to rationals, right? (This would only happen if we needed to declare a variable or a function argument with the rational type, which I think we rule out.) So to stay defensive, perhaps we should throw an error instead of just producing `double`.

src/main/scala/backends/VivadoBackend.scala
```diff
@@ -82,8 +82,10 @@ private class VivadoBackend extends CppLike {
     case _:TVoid => text("void")
     case _:TBool | _:TIndex | _:TStaticInt => text("int")
     case _:TFloat => text("float")
-    case _:TDouble => text("double")
+    case _:TDouble | _:TRational => text("double")
```
sampsyo: As above.

tissue3: Done

src/main/scala/backends/CppRunnable.scala
tissue3: Done

src/main/scala/typechecker/Subtyping.scala
tissue3: Done

src/main/scala/typechecker/TypeCheck.scala
tissue3: Done

src/main/scala/common/Syntax.scala
tissue3: Done

rachitnigam:
Can you add tests to both `TypeChekerSpec.scala` and a file based test under `src/test/should-run` that demonstrate how fixed points are used and what kinds of errors they might throw?

tissue3:
> Can you add tests to both `TypeChekerSpec.scala` and a file based test under `src/test/should-run` that demonstrate how fixed points are used and what kinds of errors they might throw?

I added both. Please check to see if that's enough.

src/main/scala/Parser.scala
```diff
@@ -121,6 +121,10 @@ private class FuseParser extends RegexParsers with PackratParsers {
     "bool" ^^ { _ => TBool() } |
     "bit" ~> angular(number) ^^ { case s => TSizedInt(s, false) } |
     "ubit" ~> angular(number) ^^ { case s => TSizedInt(s, true) } |
+    "fix" ~> angular(number~","~number ) ^^{ case s1~_~s2 => TFixed(s1,s2,false) } |
+    "fix" ~> angular(number) ^^{ case s => TFixed(s, s, false) } |
```
rachitnigam: What’s the difference between these two? Is the second one the same as a bit.

If so, either unify bit and fix or require two arguments.

src/main/scala/Parser.scala
```diff
@@ -121,6 +121,10 @@ private class FuseParser extends RegexParsers with PackratParsers {
     "bool" ^^ { _ => TBool() } |
     "bit" ~> angular(number) ^^ { case s => TSizedInt(s, false) } |
     "ubit" ~> angular(number) ^^ { case s => TSizedInt(s, true) } |
+    "fix" ~> angular(number~","~number ) ^^{ case s1~_~s2 => TFixed(s1,s2,false) } |
+    "fix" ~> angular(number) ^^{ case s => TFixed(s, s, false) } | 
+    "ufix" ~> angular(number~","~number ) ^^{ case s1~_~s2 => TFixed(s1,s2,true) } |
+    "ufix" ~> angular(number) ^^{ case s => TFixed(s, s, true) } |
```
rachitnigam: Same as above.

src/main/scala/typechecker/Subtyping.scala
```diff
@@ -57,21 +57,33 @@ object Subtyping {
...
 
   private def joinOfHelper(t1: Type, t2: Type, op: BOp): Option[Type] = (t1, t2) match {
+    //XXX(Zhijing): what happens for multiplication? Overflow?
     case (TStaticInt(v1), TStaticInt(v2)) => op.toFun match {
       case Some(fun) => Some(TStaticInt(fun(v1, v2)))
       case None => Some(TSizedInt(max(bitsNeeded(v1), bitsNeeded(v2)), false))
     }
+    case (TRational(v1), TRational(v2)) => op.toFun match {
+      //XXX(Zhijing):deprecated
+      case Some(fun) => Some(TRational(fun(v1.toDouble.toInt, v2.toDouble.toInt).toString))
```
rachitnigam: This seems wrong. op.toFun returns a function that does type level computation for naturals.

Here, you’re doing a computation by converting the rational values to integers (losing precision).

If you want to do type level computation, define a toDoubleFun function and do a precise calculation. Doing this does break the promise that rationals are infinite precision.

src/main/scala/Parser.scala
tissue3: That's the default of ap_fixed. There exists both ap_int type and ap_fixed type. If you define ap_fixed<3>, it is basically the same as ap_fixed<3,3> in HLS, but it's allowed. Just like you can say double x = 1.0, which is just the same value as int x = 1. But the type is different.

src/main/scala/Parser.scala
tissue3: Same

src/main/scala/typechecker/Subtyping.scala
tissue3: There is no way (or too troublesome) to do a precise calculation because that's the natural of floating point calculation in PC. I can either remove toInt, or just no long do any computation.

rachitnigam: If you remove it, you have to choose a type to retune for these computations. If you return double, it’ll fail for

let x: fix<3, 3> = 1.2 + 3.4

src/main/scala/Parser.scala
rachitnigam: Can we just have the first format (I.e. both Params are required). I don’t like the inconsistency of two interfaces for the fixed type. For polymorphism, it would make more sense to have one representation.

src/main/scala/typechecker/Subtyping.scala
tissue3: No. I just keep the one with larger range in it's integer part. See the None case:
if (bitsNeeded(v1.toDouble.toInt)>bitsNeeded(v2.toDouble.toInt)) Some(TRational(v1))
        else Some(TRational(v2))

src/main/scala/Parser.scala
tissue3: In fact, fixed point can require more arguments (for overflow and quantization mode), but I just haven't implemented. I don't feel any necessity to remove it just like you can define double x = 1.0.

rachitnigam: > I don't feel any necessity to remove it just like you can define double x = 1.0.

I don’t follow this.

There is a difference between the interface exposed by C++ and a Dahlia type. When you define a type T<M>, you’re describing a type level functions T that takes one parameter (M) and returns a new type. In C++, templates describe compile time computation that is more flexible than a normal type system which lets you write the two forms.

Having this means you have two function fix with different arities. Dahlia tries to expose an explicit interface to all the C++ HLS stuff which is why I want one of the two. You can represent the second form with the first one so why have two forms?

src/main/scala/typechecker/Subtyping.scala
```diff
@@ -57,21 +57,33 @@ object Subtyping {
...
     case (_:TStaticInt, _:TIndex) => true
     case (TArray(tsub, subDims), TArray(tsup, supDims)) => {
       // Arrays are invariant
       areEqual(tsup, tsub) && subDims == supDims
     }
     case (_:TFloat, _:TDouble) => true
+    case (_:TRational, _:TFloat) => true
+    case (_:TRational, _:TDouble) => true
+    case (TRational(v1), TFixed(_,i2,un2)) => 
+      ( (v1.toDouble<0 && un2==false) | (v1.toDouble>=0) ) && bitsNeeded(v1.toDouble.toInt)<=i2
```
rachitnigam: This should be `||` instead of `|`.

src/main/scala/typechecker/Subtyping.scala
rachitnigam: Maybe I'm not following what this code is supposed to do. Can you explain what should happen here?

The `Some` case triggers for all arithmetic operators.
The `None` case only triggers when you compare two numbers. (Look at the calls to `joinOf` in TypeCheck.scala)

rachitnigam:
Two tests the demonstrate the issues I'm talking about:

1. `let x: fix<3, 2> = 1.0; y < 10.0` which erroneously throws an error.
2. `let x: fix<2, 2> = 2.5 + 1.5` which doesn't throw a size check error like `let x: fix<2, 2> = 4.0` does.

rachitnigam:
This looks good! Remove the comments and hit the green button!

PR #259, opened 2019-10-02
Support multi ported memories

rachitnigam:
- Syntax: `bit<32>{2}[10 bank 5][10 bank 2]` creates a memory where each bank has two ports.


rachitnigam:
Should not be generating resource pragma for interface arrays.

PR #264, opened 2019-10-18
Origin/loopcheck

tissue3:
Check loop dependency.


src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+import CompilerError._
+import Checker._
+import EnvHelpers._
+
+object LoopChecker {
+
+  def check(p: Prog) = LCheck.check(p)
+  object States extends Enumeration{
+    type States = Value
+    val USE, DEF, DK = Value
```
rachitnigam: Scala enums should use CamelCase: `Use`, `Def`, `DontKnow`.

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
+package fuselang.passes
+
+import scala.{PartialFunction => PF}
+
+//import fuselang.Utils.RichOption
```
rachitnigam: Delete if not needed.

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+object LoopChecker {
+
+  def check(p: Prog) = LCheck.check(p)
+  object States extends Enumeration{
+    type States = Value
+    val USE, DEF, DK = Value
+  }
+  import States._
+
+  // Bounds checker environment doesn't need to track any information. Empty
```
rachitnigam: Copy pasta comment. Should be LoopCheck.

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+  // environment that just runs the commands.
+  private case class LEnv(
+    StateMap: ScopedMapExt[Id, States] = ScopedMapExt(),
+    NameMap: ScopedMap[Id,Id] = ScopedMap())(implicit val res: Int) 
+    extends ScopeManager[LEnv] {
+    
+    
+    // Helper functions for NameMap
+    def addName(vid:Id, tid:Id):LEnv = NameMap.add(vid,tid) match{
```
rachitnigam: Minor nit: There should be a space after the colon: `vid: Id`.

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+      outerenv
+    }
+    
+    def mergeHelper(k: Id, v1: Option[States], v2: Option[States], env: LEnv): LEnv = (v1, v2) match {
+      case (None, None) => throw Impossible("No such merging")
+      case (None, Some(USE)) =>  env.copy(StateMap = env.StateMap + (k->USE) )
+      case (None, _) =>  env.copy(StateMap = env.StateMap + (k->DK) )
+      case (Some(DEF), Some(DK)) =>  env.copy(StateMap = env.StateMap + (k->DK) )
+      case (Some(DEF), Some(USE)) => throw LoopDepMerge(k)
+      case (Some(DK), Some(USE)) =>  throw LoopDepMerge(k)
```
rachitnigam: These two cases can be combined using `|`.

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+        val e1 = checkE(cond)(e)
+        e1.withScope(1)(env => checkC(body)(env))
+      }
+      case (CIf(cond, c1, c2), e) => {
+        e.withScope( someScope =>{
+          val nEnv = checkE(cond)(someScope) 
+          val e1 = nEnv.withScope(checkC(c1)(_))
+          val e2 = nEnv.withScope(checkC(c2)(_))
+          val eres = e1 merge e2
+          println(e1.StateMap)
```
rachitnigam: Remove `println`.

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  private final case object LCheck extends PartialChecker {
+
+    type Env = LEnv
+
+    val emptyEnv = LEnv()(1)
+
+    def myCheckLVal(e: Expr, env: Env): Env = { e match
```
rachitnigam: Don't need the starting `{`. You can just write:
```
def myCheckLVal(e: Expr, env: Env): Env = e match {
  case ... 
```

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+      case None | Some(DK) => LEnv(StateMap+(id -> DEF), NameMap)
+      case Some(DEF) => this 
+      case Some(USE) => throw LoopDepSequential(id)
+      case _ => throw Impossible("No such state")
+    }
+    def atDk(id:Id):LEnv = StateMap.get(id) match{
+      case None | Some(DEF) => LEnv(StateMap+(id -> DK), NameMap)
+      case Some(DK) => this
+      case Some(USE) => throw LoopDepSequential(id) 
+      case _ => throw Impossible("No such state")
```
rachitnigam: Why do you need the `case _`? It seems like you're matching all possible enum cases.

If this is because of exhaustiveness checking warning, you can use `case object`s instead of an enumeration:
```
sealed trait States
case object Def extends States
case object Use extends States
case object DontKnow extends States
```

src/main/scala/common/ScopeMapExt.scala
```diff
@@ -0,0 +1,66 @@
+package fuselang.common
+
```
rachitnigam: Can you describe the changes you needed to make for `ScopeMapExt`. I'm pretty sure we can merge it with `ScopeMape`.

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+    }
+    def atUse(id:Id):LEnv = StateMap.get(id) match {
+      case None => LEnv(StateMap+(id -> USE), NameMap) 
+      case Some(DK) => throw LoopDepSequential(id)
+      case Some(DEF)| Some(USE) => this //Use/Def -> Use don't update
+      case _ => throw Impossible("No such state")
+    }
+    //check and update the state table
+    def updateState(id:Id, state: States):LEnv = {
```
rachitnigam: Seems like `updateState` and the various `at*` functions can be combined if you match using:
```
(state, StateMap.get(id)) match {
  case (Use, None) => ...
```

src/main/scala/passes/LoopCheck.scala
```diff
@@ -0,0 +1,174 @@
...
+  object States extends Enumeration{
+    type States = Value
+    val USE, DEF, DK = Value
+  }
+  import States._
+
+  // Bounds checker environment doesn't need to track any information. Empty
+  // environment that just runs the commands.
+  private case class LEnv(
+    StateMap: ScopedMapExt[Id, States] = ScopedMapExt(),
```
rachitnigam: Field names and variable names in scale should [start with lowercase](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods).

rachitnigam:
@tissue3 any reason to not merge this? Do we want a more precise analysis that tries to disambiguate reads and writes to `a[i]` and `a[i + 1]` or something?

tissue3:
Two issues:
1. indexing, i.e, view v_a = a[i] only trace the ith instance of list a, but it is fine now since no benchmark failed because of this.
2. pointers as inputs to functions. Currently loop checker assumes all input argument of a function are not lhs value and would not be modified. However, an array `arr` is passed as input argument of a function, `f(arr)` might have changed `arr`. In this example, it seems this kind of operation should be forbidden in an unrolled loop since it involves write and read in the same loop.

src/main/scala/common/Errors.scala
```diff
@@ -83,6 +83,13 @@ object Errors {
   case class AlreadyBound(id: Id) extends TypeError(
     s"`$id' already bound in scope", id.pos)
 
+  // Binding errors
+  case class LoopDepSequential(id: Id) extends TypeError(
+    s"`$id' cannot be used and then defined in an unrolled loop.", id.pos)
+
+  case class LoopDepMerge(id: Id) extends TypeError(
+    s"`$id' cannot be merged with conflit state. The execution order is non-deterministic.", id.pos)
```
sampsyo:
```suggestion
    s"`$id' cannot be merged with conflicting state. The execution order is non-deterministic.", id.pos)
```

sampsyo:
> indexing, i.e, view v_a = a[i] only trace the ith instance of list a, but it is fine now since no benchmark failed because of this.

To clarify, by "the ith instance of list a," it sounds like you're saying that `a[1]` is tracked as a different location from `a[2]`, but I don't think that's correct—right? That is, `view v_a = a[i:]`, for example, will just make `v_a` "count" the same as `a`, AFAICT. Am I understanding correctly?

> pointers as inputs to functions

Yes, this is a good point! In fact, we have plenty of soundness holes along theses lines with respect to interprocedural analysis. For example, consider:

```
def f(a: int[10 bank 2], x: int) {
  a[0] := x;
}

[...]

for (let i = 0..10) unroll 2 {
  f(b, i);
}
```

The assignment wouldn't be allowed in the unrolled context, but by "sneaking" it into a function, we can avoid the error.

Addressing this kind of interprocedural problem requires some deeper thinking, beyond just this question of eliminating cross-iteration dependencies. So I think it would be OK to merge this for now and come back to that larger topic later.

tissue3:
> > indexing, i.e, view v_a = a[i] only trace the ith instance of list a, but it is fine now since no benchmark failed because of this.
> 
> To clarify, by "the ith instance of list a," it sounds like you're saying that `a[1]` is tracked as a different location from `a[2]`, but I don't think that's correct—right? That is, `view v_a = a[i:]`, for example, will just make `v_a` "count" the same as `a`, AFAICT. Am I understanding correctly?
Yes, exactly. The issue is `a[i+2] = a[i]+1` should be allowed to unroll by 2, but I haven't done any indexing analysis.

sampsyo:
Great. Precisely analyzing array indices as separate locations seems very hard to me, so that is worth avoiding, I think, until we absolutely can't get around it anymore.

PR #279, opened 2019-11-14
fix loop dependency checker, thanks to sam

tissue3:


rachitnigam:
@tissue3 can you explain what the issue was and what the fix was?

PR #280, opened 2019-11-14
fixed loop checker

tissue3:
This is the true loopfix. There is something wrong with nameMap in the last version.

NameMap need addScope and endScope methods within both `if` scopes and loop scopes (`while` and `for`). StateMap does not for `if` statements, but for loops it does.
That makes both last pull request of fixing (which does not create scope for nameMap at an `if` statement) and the old loop checker both wrong (which create scope for both nameMap and stateMap at an `if` statement).

### DIFF
--- a/src/main/scala/passes/LoopCheck.scala
+++ b/src/main/scala/passes/LoopCheck.scala
@@ -71,6 +71,18 @@ object LoopChecker {
         case env:LEnv => env.endScope(resources)
       }
     }
+    def withScope(loop: Boolean, resources: Int)(inScope: LEnv => LEnv): LEnv = {
+      if (loop == false){
+          inScope(this.addIfScope) match {
+            case env:LEnv => env.endIfScope
+          }
+      }else{
+          inScope(this.addScope(resources)) match {
+            case env:LEnv => env.endScope(resources)
+          }
+      }
+    }
+    
     // To satisfy envhelper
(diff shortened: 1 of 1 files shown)


rachitnigam:
This doesn't explain the issue @sgpthomas discovered and what the checker was doing on it.

tissue3:
> This doesn't explain the issue @sgpthomas discovered and what the checker was doing on it.

It does. In the version @sgpthomas finds wrong, loop checker add scope for both stateMap and nameMap at an `if` statement. Actually, for `if` statement, we just need to add scopoe for namMap, but not stateMap. StateMap should call add scope only when there is loop statements (`while` and `for`)

rachitnigam:
My question is why is adding a scope for `if` in `stateMap` wrong? I understand the code changes. I want to get an intuition for the fix.

tissue3:
Try to get an intuition of why we need a new `stateMap` for a loop. It was because the loop can have back-edges and `use-def` chain could goes wrong since when we try to `use` a variable `x`, we don't know the `def` of `x` comes from last loop or before the loop starts. It is not a problem for `if` statement though. We use `x` in `if`, then it's definition must come from outside of the loop. That is, `def-use` chain is stable and does not need any special operations on it.

tissue3:
Please don't merge this PR for now. I just realized a better way to arrange the code.

tissue3:
Done

PR #284, opened 2019-11-15
quick fix to indexing of loop checker

tissue3:
Now if the index is completely the same at the dimension requires unrolling, loop checker says there is not have cross loop dependency, but it is not smart enough to check view.

### DIFF
--- a/src/test/scala/TypeCheckerSpec.scala
+++ b/src/test/scala/TypeCheckerSpec.scala
@@ -793,9 +793,9 @@ class TypeCheckerSpec extends FunSpec {
     it("use after define in an unrolled loop is not allowed") {
       assertThrows[LoopDepSequential] {
         typeCheck("""
-          decl a: bit<32>[10 bank 5];
-          for (let i = 0..10) unroll 5 {
-            let x = a[i]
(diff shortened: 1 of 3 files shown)


src/test/scala/TypeCheckerSpec.scala
```diff
@@ -793,9 +793,9 @@ class TypeCheckerSpec extends FunSpec {
     it("use after define in an unrolled loop is not allowed") {
       assertThrows[LoopDepSequential] {
         typeCheck("""
-          decl a: bit<32>[10 bank 5];
-          for (let i = 0..10) unroll 5 {
-            let x = a[i]
+          let a: bit<32>{2}[10];
+          for (let i = 0..10) unroll 2 {
+            let x = a[i+1]
```
rachitnigam: Why did this test need to be changed?

rachitnigam:
Can you add some tests showing this behavior?

src/test/scala/TypeCheckerSpec.scala
tissue3: originally the test the code was basically `a[i]=a[i]+1`, which is no longer a loop dependency error.

PR #285, opened 2019-11-18
fix scoping for dependent loops

sgpthomas:
I wasn't able to get the ScopedSets working correctly, so I just made a much simpler change. Writing this I realized that this doesn't work in all cases.

this throws an error but shouldn't
```c
let foo: float[2];
for (let j = 0..2) unroll 2 {
  let y: bit<100> = j + 1;
  for (let i = 0..2) unroll 2 {
    y := 3;
  }
  for (let i = 0..2) unroll 2 {
    foo[y];
  }
}
```


rachitnigam:
@sgpthomas what is this branch for?

sgpthomas:
tbh, I don't quite remember. We should rewrite most of this code anyways. I'm going to delete this branch

PR #288, opened 2019-11-22
Loop flatten

sgpthomas:


rachitnigam:
@sgpthomas what is this branch for?

sgpthomas:
This was a hack to get hardware estimates for one of the kernels. It turned out that unless you explicitly turned off `loop flattening` in Vivado, it would flatten loops and get rid of the trip count pragmas and then fail to generate hardware estimations. We should make a more informed decision about what to do about Vivado automatically applying particular optimizations so that we can get more predictable results from Vivado.

rachitnigam:
It turned out the HLS responds differently when you explicitly say `unroll=1` and leave out the `unroll` instruction.

[stencil_vivado_hls_with_explicit.log](https://github.com/cucapra/dahlia/files/4267919/stencil_vivado_hls_with_explicit.log)
[stencil_vivado_hls_without_explicit.log](https://github.com/cucapra/dahlia/files/4267920/stencil_vivado_hls_without_explicit.log)

PR #267, opened 2019-10-24
Drew refactor

dz333:
This is for hacktoberfest issue #265 
Very minimal work has been done. But it does work.

PR #290, opened 2020-02-09
Separate Affine checker

### LINKED ISSUE #262: Wrong merge function at combine
Dahlia does not allow the following code:
```
decl m1: float[32][32 bank 32];
decl m2: float[32 bank 32][32];
decl prod: float[32][32];

for (let i = 0..32) {
  for (let j = 0..32) {
    for(let k = 0..32) unroll 32 {
      let mult = m1[i][k] * m2[k][j];
      ---
      m1[i][k] := 0.0;
    } combine {
      prod[i][j] += mult;
    }
  }
}
```
 It will reports `mult` is not bound in scope. It's wrong. `mult` is bounded in scope. 
The issue is the `merge` function on type environments doesn't correctly merge the sets of bound identifiers according to @rachitnigam .

### LINKED ISSUE #265: Separate well formedness and affine checking from typechecker
The current typechecker is written as a single monolithic function that does simple typechecking (check operands to + are numbers etc.), affine resource tracking (memory read/writes don't conflict), and well formedness checks (`let` bound arrays and record literals have types).

This issue to separate these things into three distinct passes. Separating well formedness from type checking should be easier than separating affine checking.

The order of passes should be:

Parser > Well Formedness > Type Check > Affine Check > ....

rachitnigam:
Fixes #203. Fixes #262. Fixes #265.

Three large changes:
- Separate out normal type checking from affine type checking.
- Separate out well formedness checks from type checking. Subsumes #267.
- Reduction operators operate on normal scalars and are disabled inside parallel parts of for loops.

------

**TODO**
Test behavior on:
- [X] Machsuite rewrites
- [X] DSE: md-grid
- [x] DSE: md-knn
- [x] DSE: stencil2d


rachitnigam:
All DSE experiments accept the same set of programs. Cleanup and merge!

PR #300, opened 2020-07-09
Generate assignments to done holes

sgpthomas:
fixed backend to generate done hole assignments for groups and generate correct bitwidths for components

### DIFF
--- a/src/main/scala/backends/futil/FutilAst.scala
+++ b/src/main/scala/backends/futil/FutilAst.scala
@@ -167,8 +167,8 @@ object Futil {
   sealed trait GuardExpr extends Emitable {
     override def doc(): Doc = this match {
       case Atom(item) => item.doc
-      case And(left, right) => left.doc <+> text("&") <+> right.doc
-      case Or(left, right) => left.doc <+> text("|") <+> right.doc
+      case And(left, right) => parens(left.doc <+> text("&") <+> right.doc)
+      case Or(left, right) => parens(left.doc <+> text("|") <+> right.doc)
     }
   }
 
--- a/src/main/scala/backends/futil/FutilAst.scala
+++ b/src/main/scala/backends/futil/FutilAst.scala
@@ -74,18 +74,31 @@ object Futil {
       case CompPort(id, name) =>
         id.doc <> dot <> text(name)
       case ThisPort(id) => id.doc
+      case HolePort(id, name) =>
+        id.doc <> brackets(text(name))
+      case ConstantPort(width, value) =>
+        text(width.toString) <> text("'d") <> text(value.toString)
     }
 
     override def compare(that: Port): Int = (this, that) match {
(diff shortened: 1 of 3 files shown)


src/main/scala/backends/futil/FutilAst.scala
```diff
@@ -148,6 +164,18 @@ object Futil {
     }
   }
 
+  sealed trait GuardExpr extends Emitable {
+    override def doc(): Doc = this match {
+      case Atom(item) => item.doc
+      case And(left, right) => left.doc <+> text("&") <+> right.doc
+      case Or(left, right) => left.doc <+> text("|") <+> right.doc
```
rachitnigam, line 171: Do you not need to parenthesize this?

src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -7,6 +7,17 @@ import Syntax._
 import Configuration._
 import CompilerError._
 
+/** Helper class that gives names to the fields of the output of `emitExpr` and `emitBinop`.
+  *  - `port` holds either an input or output port that represents how data flows between exprs
+  *  - `done` holds the port that signals when the writing or reading from `port` is done
+  *  - `structure` represents additional structure involved in computing the expression
+  */
+private class EmitOutput(
```
rachitnigam, line 15: Probably want to make this a case class so you don't need to call `new` down below.

src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -121,33 +156,50 @@ private class FutilBackendHelper {
...
           }
         })
 
         // The value is generated on `read_data` and written on `write_data`.
         val portName = if (lhs) "write_data" else "read_data"
-        (arr.port(portName), indexing)
+        val writeEnStruct =
+          if (lhs) List(Connect(ConstantPort(1, 1), arr.port("write_en")))
+          else List()
+        new EmitOutput(
```
rachitnigam, line 198: Remove `new` when EmitOutput is a case class.

src/main/scala/typechecker/TypeCheck.scala
```diff
@@ -131,7 +131,7 @@ object TypeChecker {
     case EArrLiteral(_) => throw NotInBinder(expr.pos, "Array Literal")
     case ECast(e, castType) => {
       val (typ, nEnv) = checkE(e)
-      if (safeCast(typ, castType) == false) {
+      if (!safeCast(typ, castType)) {
```
rachitnigam, line 134: I've always preferred explicit `== false` instead of `!` which is easy to miss. Also, change not needed in this commit.

src/main/scala/backends/futil/FutilAst.scala
sgpthomas: Nope, this was just an oversight on my part. thanks for catching this!

PR #302, opened 2020-07-17
CppRunable prints out a JSON

rachitnigam:


src/main/scala/backends/CppRunnable.scala
```diff
@@ -36,7 +38,19 @@ private class CppRunnable extends CppLike {
     case TAlias(n) => value(n)
   }
 
-  def emitArrayDecl(ta: TArray, id: Id) = emitType(ta) <+> text(s"$id")
+  def emitArrayDecl(ta: TArray, id: Id) =
+    emitType(ta) <+> text("*") <> text(s"$id")
+
+  override def emitDecl(id: Id, typ: Type): Doc = {
```
rachitnigam, line 47: @sgpthomas Why are you doing this?

PR #305, opened 2020-07-31
Vivado Customizable Mem Interfaces

sgpthomas:
- Add a `--memory-interface` flag to customize the memory interfaces that Dahlia generates for Vivado.
 - Change the Futil backend to emit external memories for `decls`

### DIFF
--- a/src/main/scala/backends/futil/FutilBackend.scala
+++ b/src/main/scala/backends/futil/FutilBackend.scala
@@ -51,6 +51,10 @@ private class FutilBackendHelper {
     */
   type Store = Map[CompVar, CompVar]
 
+  /** `external` is a flag that differentiates between generating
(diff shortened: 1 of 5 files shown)


src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -51,7 +51,7 @@ private class FutilBackendHelper {
     */
   type Store = Map[CompVar, CompVar]
 
-  def emitArrayDecl(typ: TArray, id: Id): List[Structure] = {
+  def emitArrayDecl(typ: TArray, id: Id, external: Boolean): List[Structure] = {
```
rachitnigam, line 54: Comment describing what `external` does here?

rachitnigam:
One comment needs resolving. LGTM apart from that. Merge when done.

PR #308, opened 2020-08-05
[WIP] Lower unroll and bank

rachitnigam:
Rewrite unrolled `for` loops and `bank` memories.

**TODO**
- [x] Support dynamic read and writes for banked memories
- [ ] Add tests

PR #309, opened 2020-08-07
add a pass that buffers read and writes to memories

sgpthomas:


rachitnigam:
#308 already implemented a pass to register reads (stolen from this PR). If you don't need write registering, close this.

PR #315, opened 2020-08-14
Hoist binop

sgpthomas:
This implements a Dahlia pass that breaks up Futil binary expressions that are multi cycle. The changes to the Futil backend are minimal. The hoisted code is more sequential than it needs to be, but I'm going to leave fixing that to a later point.


src/main/scala/Compiler.scala
```diff
@@ -16,15 +16,17 @@ object Compiler {
     "Hoist memory reads" -> passes.HoistMemoryReads,
     "Sequentialize" -> passes.Sequentialize,
     "Lower unroll and bank" -> passes.LowerUnroll,
-    "Lower for loops" -> passes.LowerForLoops
+    "Lower for loops" -> passes.LowerForLoops,
+    "Hoist slow binops" -> passes.HoistSlowBinop
```
rachitnigam, line 20: Why not run this after hoist memory reads? Is it because `lower unroll` might generate expense expressions?

src/main/scala/Compiler.scala
```diff
@@ -16,15 +16,17 @@ object Compiler {
...
   // pass should only run during lowering.
-  val postTransformers: List[(String, (TypedPartialTransformer, Boolean))] = List(
-    "Rewrite views" -> (passes.RewriteView, false),
-    "Add bitwidth" -> (passes.AddBitWidth, true)
-  )
+  val postTransformers: List[(String, (TypedPartialTransformer, Boolean))] =
+    List(
+      "Rewrite views" -> (passes.RewriteView, false),
+      "Add bitwidth" -> (passes.AddBitWidth, true)
+    )
```
rachitnigam, line 29: When submitting PRs, aspire to minimize the diffs so that they don't show unnecessary changes. If you're going to reformat a whole file, commit the formatting changes separately (this is standard practice at every software company)

src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -188,7 +188,50 @@ private class FutilBackendHelper {
       comp.id.port("out"),
       ConstantPort(1, 1),
       struct ++ e1Out.structure ++ e2Out.structure,
-      Some(0)
+      for (d1 <- e1Out.delay; d2 <- e2Out.delay)
```
rachitnigam, line 191: Is `delay` and `Option[T]`? Why? Everything should have a delay.

rachitnigam: I see the problem below.

src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -339,13 +393,17 @@ private class FutilBackendHelper {
...
+            sqrt.id.port("go"),
+            Some(Not(Atom(sqrt.id.port("done"))))
+          )
         )
         EmitOutput(
           sqrt.id.port("out"),
           sqrt.id.port("done"),
           argOut.structure ++ struct,
-          Some(1)
+          Some(17)
```
rachitnigam, line 406: No magic numbers allowed.

src/main/scala/passes/HoistSlowBinop.scala
```diff
@@ -0,0 +1,122 @@
+package fuselang.passes
+
+import scala.{PartialFunction => PF}
+import fuselang.common._
+import Syntax._
+// import CodeGenHelpers._
+// import CompilerError._
```
rachitnigam, line 7: Remove the commented imports

src/main/scala/passes/HoistSlowBinop.scala
```diff
@@ -0,0 +1,122 @@
...
+  def binopRecur(expr: Expr, env: Env): (Expr, Env) = {
+    expr match {
+      // only recur when children are binops
+      case EBinop(_, _, _) => rewriteE(expr)(env)
+      case _ => (expr, env)
+    }
+  }
+
+  def myRewriteE: PF[(Expr, Env), (Expr, Env)] = {
+    case (e @ EBinop(op, left, right), env) if slowBinops.contains(op.op) => {
```
rachitnigam, line 61: A large part of this code overlaps with the HoistMemoryReads traversal (unsurprisingly). Is there a way to abstract this pattern into a class that can be instantiated with a `Expr => boolean` function and hoists everything that it gets `true` for?

src/main/scala/passes/HoistSlowBinop.scala
```diff
@@ -0,0 +1,122 @@
...
+        env
+      )
+    }
+    case (CUpdate(lhs, rhs), _) => {
+      val (rewrLhs, env) = rewriteE(lhs)(emptyEnv)
+      val (rewrRhs, nEnv) = rewriteE(rhs)(env)
+      construct(CUpdate(rewrLhs, rewrRhs), nEnv)
+    }
+    // XXX(sam): I probably should deal with `*=`
+    case (CReduce(rop, lhs, rhs), _) => {
```
rachitnigam, line 103: If your code doesn't deal with it, `throw` an error. I'd much rather have errors than silently generated bad code.

src/main/scala/Compiler.scala
sgpthomas: yeah. I haven't actually verified this though. do you think it should be run before unroll?

src/main/scala/backends/futil/FutilBackend.scala
sgpthomas: this is just the static time of the sqrt component. I'll leave a comment

src/main/scala/Compiler.scala
rachitnigam: Unrolling "just" works with hoisting since it doesn't have any assumptions about the structure apart from "all memory reads are registered".

The only problem is that shrink views might generate `%` and `/` when the bank number is not a power of two. I think as long as you are not introducing too many `---`, it should be fine. I guess there needs to some `---` between the computation of the multiply and the use of the register.

sgpthomas:
I agree that abstracting the traversal would be nice, but I'm not going to prioritize it now. I'll open an issue for it.

src/main/scala/backends/futil/FutilBackend.scala
rachitnigam: Can you make a `Map` or something and add a key-value pair to it. I assume mult also has a similarly hard coded number somewhere.

rachitnigam:
Sure

> *He who doesn't abstract shall suffer the wrath of duplicated bugs*

src/main/scala/passes/HoistSlowBinop.scala
```diff
@@ -99,8 +98,10 @@ object HoistSlowBinop extends TypedPartialTransformer {
       val (rewrRhs, nEnv) = rewriteE(rhs)(env)
       construct(CUpdate(rewrLhs, rewrRhs), nEnv)
     }
-    // XXX(sam): I probably should deal with `*=`
     case (CReduce(rop, lhs, rhs), _) => {
+      rop.op match {
+        case "*=" | "/=" => throw NotImplemented(s"Hoisting $rop.op")
```
rachitnigam, line 103: NotImplemented can take a location. Pass the location of the `op`.

PR #317, opened 2020-08-15
Smart Sequentialize

rachitnigam:
Slightly smarter sequentialize pass that doesn't add `---` when the commands use different variables


rachitnigam:
@sgpthomas can you try merging this (making sure the tests all pass). I think it works but haven't tested with all FuTIL tests.

PR #324, opened 2020-11-01
Allow for custom component name when lowering to FuTIL.

cgyurgyik:
This passes in the kernel name of the configuration instead of a default string `"main"`.
So, when 
```
 ./fuse ./../futil/examples/dahlia/vectorized-add.fuse -n NewName -b futil --lower
```
is called, it produces: 
```
...
component NewName() -> () { ... }
``` 

However, it also means that the default would be "kernel" rather than "main," so what should we do here?

### DIFF
--- a/src/main/scala/backends/futil/FutilBackend.scala
+++ b/src/main/scala/backends/futil/FutilBackend.scala
@@ -651,7 +651,7 @@ private class FutilBackendHelper {
       "prog",
       List(
         Import("primitives/std.lib"),
-        Component(c.KernelName, List(), List(), struct.sorted, control)
+        Component(c.kernelName, List(), List(), struct.sorted, control)
       )
     ).emit
   }
--- a/src/main/scala/backends/futil/FutilBackend.scala
+++ b/src/main/scala/backends/futil/FutilBackend.scala
@@ -651,7 +651,7 @@ private class FutilBackendHelper {
       "prog",
       List(
         Import("primitives/std.lib"),
-        Component("main", List(), List(), struct.sorted, control)
+        Component(c.KernelName, List(), List(), struct.sorted, control)
       )
     ).emit
   }
--- a/src/main/scala/backends/futil/FutilBackend.scala
+++ b/src/main/scala/backends/futil/FutilBackend.scala
@@ -651,7 +651,7 @@ private class FutilBackendHelper {
       "prog",
       List(
         Import("primitives/std.lib"),
-        Component(c.kernelName, List(), List(), struct.sorted, control)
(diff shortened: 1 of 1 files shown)


sampsyo:
That's a good question! The hacky solution, which isn't so bad, would be to make the default different for C++ vs. FuTIL… i.e., the configuration would indicate whether we're in the default state, and the backend would use that to choose `kernel` (in `CppBackend`) or `main` (in `FutilBackend`).

I guess the alternative is to change Fud to invoke the Dahlia compiler with an explicit function name, but that seems like more hassle than it's worth.

cgyurgyik:
> That's a good question! The hacky solution, which isn't so bad, would be to make the default different for C++ vs. FuTIL… i.e., the configuration would indicate whether we're in the default state, and the backend would use that to choose `kernel` (in `CppBackend`) or `main` (in `FutilBackend`).
> 
> I guess the alternative is to change Fud to invoke the Dahlia compiler with an explicit function name, but that seems like more hassle than it's worth.

I'm not familiar at all with Scala, so did something even worse...

sampsyo:
Seems legit! Ready for an @sgpthomas review, perhaps?

src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -651,7 +651,7 @@ private class FutilBackendHelper {
       "prog",
       List(
         Import("primitives/std.lib"),
-        Component("main", List(), List(), struct.sorted, control)
+        Component(if (c.kernelName == "kernel") "main" else c.kernelName, List(), List(), struct.sorted, control)
```
sgpthomas, line 654: Hmm I dont like doing this. Does anything break in FuTIL if we just let `kernel` be the default name for components?

sgpthomas (changes requested):

src/main/scala/backends/futil/FutilBackend.scala
sampsyo: @sgpthomas, as I noted in https://github.com/cucapra/dahlia/pull/324#issuecomment-720163384, I think we would have to change all the FuTIL tests that use Dahlia, and maybe fud itself, to use the right name?

sgpthomas: when I get back to a computer I'll test this. But I don't think much needs to change. I think we would just have to change the Dahlia frontend tests and then the name that the testbench looks for

sampsyo: Sounds legit.

sgpthomas: yeah, we would have to change a lot of tests. I guess that we can do this for now.

sgpthomas (approved):

PR #326, opened 2020-11-02
extend diff bit

yoonachang:
still in process of testing but backend should use different primitive function for different bit additions


sampsyo:
Awesome! I know you marked this as "still in progress," but I'll be the person who brings up that a Dahlia-side test case would be awesome to have here eventually. :smiley:

PR #330, opened 2020-11-14
signed operation extended

yoonachang:


src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -209,7 +216,10 @@ private class FutilBackendHelper {
         val frac_bit1 = e1Bits - int_bit1
         val frac_bit2 = e2Bits - int_bit2
         val out_bit = int_bit1 + frac_bit2
-        val binop = if (compName =="add" && frac_bit1!=frac_bit2 ) 
+        val binop = if (compName =="add" && frac_bit1!=frac_bit2 &&(signed(e1.typ)==true))
```
rachitnigam, line 219: Space between `==` and no need for `== true`.

PR #331, opened 2020-11-20
Add D4 memory to FuTIL backend.

cgyurgyik:
This is the simplest way to currently add another dimension to FuTIL backend with the given setup. This will suffice for the VGG Net example, but as discussed, next steps are using row major lookup for scalability purposes.

I've confirm this lowers (with the FuTIL additions to the primitive library):
```
decl A: ubit<32>[8][8][8][8];
```
->
```
import "primitives/std.lib";
component main() -> () {
  cells {
    A0_0_0_0 = prim std_mem_d4_ext(32,8,8,8,8,4,4,4,4);
  }
  wires {
  }
  control {
    empty
  }
}
```


rachitnigam:
LGTM!

PR #336, opened 2020-12-06
Use FastParse for parsing

### LINKED ISSUE #299: [Performance] Rewrite parser
After a random evening of profiling things trying to see if the Dahlia compiler compiled with [scala-native](https://scala-native.readthedocs.io/en/v0.3.9-docs/) would run any faster (it was slower), it turns out the compiler spends the following % of time in the parser:

- JVM: 45-51%
- Native: 77-96%

Probably worth rewriting the compiler using [Fastparse](https://www.lihaoyi.com/fastparse/) or [Parboiled](https://github.com/sirthias/parboiled) which are reported 100x faster.

Of course, the biggest performance win would be to remove the JVM startup time which might be possible with scala native when we rewrite the parser.

rachitnigam:
Fixes #299.


src/main/scala/Parser.scala
```diff
@@ -261,8 +261,8 @@ private class FuseParser extends RegexParsers with PackratParsers {
     }
   }
   lazy val funcDef: P[FuncDef] = positioned {
-    "def" ~> iden ~ parens(repsep(args, ",")) ~ retTyp ~ block ^^ {
-      case fn ~ args ~ ret ~ body => FuncDef(fn, args, ret, Some(body))
+    "def" ~> iden ~ parens(repsep(args, ",")) ~ retTyp ~ "=" ~ block ^^ {
```
sampsyo, line 264: Any particular reason for this syntax change, adding a `=` to function definitions?

sampsyo (approved):
Awesome!! Kind of amazing that the tests pass already just with some judicious sprinkling of semicolons, right?

> I'll never touch a parser again for as long as I live.

— @rachitnigam, November 2020

src/main/scala/Parser.scala
rachitnigam: Turns out the grammar is ambiguous/requires a lot of back tracking without it. This is because code blocks have the same syntax:

```
let x = 1;
{
  let y = 2 + x;
}
```

rachitnigam:
Yeah, most of it was removing ambiguity from the grammar.

> I'll never touch a parser again for as long as I live.
> — @rachitnigam, November 2020

The commits are from Jul 2020 so I win Adrian.

rachitnigam:
Feature parity with master (positional errors etc.) Test FuTIL benchmarks before merging.

PR #346, opened 2021-01-27
Implement backend-specific imports

rachitnigam:
Adds the following syntax for backend imports:
```
import c++("math.h") 
  vivado("vivado_math.h")
  futil("sqrt.futil") {
  def sqrt(in: ubit<32>): ubit<32>;
}
```


sampsyo:
Yay!!!!!!!!

PR #347, opened 2021-02-01
Use Ports for Futil backend `invoke` parameters.

cgyurgyik:


src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -637,7 +637,7 @@ private class FutilBackendHelper {
           })
           .unzip
         val parameters =
-          id2FuncDef(invokeId).args.map(decl => CompVar(decl.id.toString()))
+          id2FuncDef(invokeId).args.map(decl => ThisPort(CompVar(decl.id.toString())))
```
rachitnigam, line 640: Can you explain why `ThisPort` is needed here?

cgyurgyik: I talk about this in [futil#381](https://github.com/cucapra/futil/issues/381). Based off what is mentioned in the original `Invoke` issue, if we want to hook into output ports, we just flip the destination and source.
For example, to write to a memory `x`, we need to hook the `write_en` and `write_data` ports to the `foo` component's output ports.

```
// Treat everything as a Port
invoke comp_foo(..., x.write_en = x_write_en, x.write_data = x_write_data)(); 
```

cgyurgyik:
```
Error: FuTIL Parser:   --> a.futil:39:59
   |
39 |       invoke foo0(a_read_data=a.read_data, a_done=a.done, a.write_data=a_write_data, a.write_en=a_write_en, a.addr0=a_addr0, b=const0.out)();
   |                                                           ^---
   |
   = expected invoke_arg
```

Hmm I just made the necessary changes to the FuTIL backend, but lowering to FuTIL gives me the following. So perhaps this isn't what you had in mind?

rachitnigam: The output ports have the same syntax as input ports:
```
invoke foo(a_read_data = a.read_data)(a_write_en = a.write_en) // port goes AFTER
```

cgyurgyik: Ah ok! I originally thought that was the case but saw different syntax in the Invoke issue.

Issue #356, opened 2021-03-06
Fixed point division not using correct primitive.


cgyurgyik:
```
let x: fix<1, 1> = (1.0 as fix<1, 1>);
let y: fix<1, 1> = (2.0 as fix<1, 1>);
let z: fix<1, 1> = x / y;
```

lowers to:
```
import "primitives/std.lib";
component main() -> () {
  cells {
    bin_read0_0 = std_reg(1);
    div_pipe0 = std_sdiv_pipe(1);
    fpconst0 = fixed_p_std_const(1,1,0,1,0);
    fpconst1 = fixed_p_std_const(1,1,0,2,0);
    slice0 = std_slice(1,1);
    slice1 = std_slice(1,1);
    slice2 = std_slice(1,1);
    slice3 = std_slice(1,1);
    x_0 = std_reg(1);
    y_0 = std_reg(1);
    z_0 = std_reg(1);
  }
  wires {
    group let0<"static"=1> {
      x_0.in = slice0.out;
      x_0.write_en = 1'd1;
      let0[done] = x_0.done;
      slice0.in = fpconst0.out;
    }
    group let1<"static"=1> {
      y_0.in = slice1.out;
      y_0.write_en = 1'd1;
      let1[done] = y_0.done;
      slice1.in = fpconst1.out;
    }
    group let2<"static"=1> {
      bin_read0_0.in = slice2.out;
      bin_read0_0.write_en = 1'd1;
      let2[done] = bin_read0_0.done;
      slice2.in = div_pipe0.out;
      div_pipe0.left = x_0.out;
      div_pipe0.right = y_0.out;
      div_pipe0.go = !div_pipe0.done ? 1'd1;
    }
    group let3<"static"=1> {
      z_0.in = slice3.out;
      z_0.write_en = 1'd1;
      let3[done] = z_0.done;
      slice3.in = bin_read0_0.out;
    }
  }
  control {
    seq {
      par {
        let0;
        let1;
      }
      let2;
      let3;
    }
  }
}
```

However, we really want to use `fixed_p_std_sdiv`.

cgyurgyik:
This is a result of:

https://github.com/cucapra/dahlia/blob/7b7e16418b331723f691e8312efecdbf80f0c8ee/src/main/scala/backends/futil/FutilBackend.scala#L489-L526

It is defaulting to using `div_pipe` and `mod_pipe` for `/` and `*` respectively.

PR #357, opened 2021-03-06
Lower fixed point operators to corresponding Calyx primitive.

cgyurgyik:
Addresses #356.

### DIFF
--- a/src/main/scala/backends/futil/FutilBackend.scala
+++ b/src/main/scala/backends/futil/FutilBackend.scala
@@ -79,12 +79,22 @@ private class FutilBackendHelper {
     }
   }
 
-  /** Returns true if the given int or fixed point is signed
+  /** Returns true if the given int or fixed point is signed.
     */
   def signed(typ: Option[Type]) = {
     typ match {
       case Some(TSizedInt(_, un)) => un == false
       case Some(TFixed(_, _, un)) => un == false
+      case _ => false
+    }
+  }
+
+  /** Returns true if the type is fixed point,
+    * false otherwise.
+    */
+  def is_fixed_point(typ: Option[Type]) = {
+    typ match {
+      case Some(TFixed(_, _, _)) => true
       case _ => false
     }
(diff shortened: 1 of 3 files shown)


src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -490,15 +500,15 @@ def emitInvokeDecl(app: EApp)(implicit store: Store, id2FuncDef: FunctionMapping
         val compName = op.op match {
           case "+" => "add"
           case "-" => "sub"
-          case "*" => "mult_pipe"
-          case "/" => "div_pipe"
+          case "*" => "mult"
+          case "/" => "div"
```
rachitnigam, line 504: No, this doesn't work. 

The main `div` and `mult` primitives that can actually be turned into hardware have the suffix `_pipe` in the standard library. The fixed point primitive implemented for `div` is not synthesizable hardware. We need an equivalent `fixed_p_div_pipe` primitive to make things work.

cgyurgyik: When not fixed point, It is being emitted with the `_pipe` suffix for multi-cycle binary operations. What am I missing here?

cgyurgyik: As far as creating synthesizable components, created an issue in Calyx:
https://github.com/cucapra/calyx/issues/415
I'll get around to this.

rachitnigam: Here is the problem: `std_mult` and `std_div` are not hardware synthesizable. the `_pipe` variants are which is why Dahlia defaults to it. This is also true of the fixed point variants which use `*` and `/` operators. The correct solution is to remove `std_mult` and `std_div` altogether and just use the synthesizable variants which is what #355 and related PRs are getting at.

A short term solution is renaming the fixedpoint operators in the Calyx library to have `_pipe` in their names.

Issue #361, opened 2021-03-30
Current Fixed Point Constant Implementation parses numbers incorrectly.


cgyurgyik:
The current implementation simply parses the numbers into signed integers on each side of the `.`:
```
  case ECast(ERational(d), typ) => {
    val _ = rhsInfo
    val (width, Some(int_bit)) = bitsForType(Some(typ), expr.pos)
    val frac_bit = width - int_bit
    val lst = d.split('.')
    val v_1 = lst(0).toInt
    val v_2 = lst(1).toInt
    val fpconst =
      Cell(
        genName("fpconst"),
        Stdlib.fixed_point(width, int_bit, frac_bit, v_1, v_2),
        false
      )
    EmitOutput(
      fpconst.id.port("out"),
      ConstantPort(1, 1),
      List(fpconst),
      Some(0)
    )
  }
```

So for example,
`1.5 as ufix<1, 1>`

Will interpret this as
Integer value: 1
Fractional value: 5

and then append their bit representations (`1`, `101`). We can’t even represent `5` with a single bit. The actual fix is to interpret the fractional value `0.5` as `1`.

The goal is to deprecate `fp_const` in #359, but I think this is worth noting for the time being.

PR #362, opened 2021-03-31
Deprecate fixed point constants in Dahlia.

### LINKED ISSUE #359: Remove fixed point constants from Dahlia.
Part of [Calyx #470](https://github.com/cucapra/calyx/pull/470).

from [dahlia/FutilBackend.scala](https://github.com/cucapra/dahlia/blob/8dd37e4b26878e96b655b23e8abbab8e1581d401/src/main/scala/backends/futil/FutilBackend.scala#L585-L604):

      case ECast(ERational(d), typ) => {
        val _ = rhsInfo
        val (width, Some(int_bit)) = bitsForType(Some(typ), expr.pos)
        val frac_bit = width - int_bit
        val lst = d.split('.')
        val v_1 = lst(0).toInt
        val v_2 = lst(1).toInt
        val fpconst =
          Cell(
            genName("fpconst"),
            Stdlib.fixed_point(width, int_bit, frac_bit, v_1, v_2),
            false
          )
        EmitOutput(
          fpconst.id.port("out"),
          ConstantPort(1, 1),
          List(fpconst),
          Some(0)
        )
      }

This will be the one place that fixed point constants are dependent on. We’ll have to instead re-interpret `v_1` and `v_2` accordingly (taking into account signedness).

It can then be removed from [here](https://github.com/cucapra/dahlia/blob/8dd37e4b26878e96b655b23e8abbab8e1581d401/src/main/scala/backends/futil/FutilAst.scala#L331-L342) as well. This will break the test [here](https://github.com/cucapra/dahlia/blob/master/file-tests/should-futil/invoke-with-fixed-point.expect), which should instead use a `std_const`.

cgyurgyik:
Closes #359, #361.


file-tests/should-futil/fixed-point-constant.fuse
```diff
@@ -0,0 +1,5 @@
+let a: ufix<4, 2> = (2.25 as ufix<4, 2>); // 9
```
rachitnigam, line 1: What are the numbers after the constants supposed to be? The bitwidth representation? If so, can you add a comment at the top of the test file saying that it is?

rachitnigam:
LGTM with a minor comment! Merge whenever.

PR #367, opened 2021-04-17
Infer SV parameters from return type in Calyx backend.

cgyurgyik:
This is a quick-fix to support SystemVerilog modules that require one or more parameters. This wasn't necessary before because no function definition utilized up to this point needed a parameter.

For example,
```
module sqrt #( parameter WIDTH )( ...);
```
requires a `WIDTH` parameter, therefore an argument must be passed in the `CompInst`. If the function ID is within a list of function IDs that "require arguments", then get the `WIDTH`(s) based off the return type.

Thus,
```s = sqrt();```
becomes
``` s = sqrt(32); // WIDTH = 32```

### CODE REVIEW

File: src/main/scala/backends/futil/FutilBackend.scala [hunk 1]
```diff
@@ -244,13 +244,43 @@ private class FutilBackendHelper {
...
+  *   module foo #(
+  *     parameter WIDTH
+  *   ) ( ... );
+  * ```
+  * Requires that a `WIDTH` be provided. Currently, the functions that
+  * do require these parameters must be manually added to the list
+  * `requiresWidthArguments`. */
+def getCompInstArgs(funcId: Id)(implicit id2FuncDef: FunctionMapping): List[Int] = {
+  // A list of functions that require width arguments.
+  val requiresWidthArguments = List("fp_sqrt", "sqrt")
```
rachitnigam, line 261: Make this a file global instead of a function local value.

rachitnigam:
LGTM with small change.

PR #368, opened 2021-04-18
Emit correct guard for `write_en` in multi-cycle groups.

### LINKED ISSUE #369: Multi-cycle binary operations emit incorrect `write_en` guard.
When lowering a multiply to Calyx, we get a group such as the one below:
```
    group mult<"static"=1> {
      r.in = mult_pipe0.out;
      r.write_en = 1'd1;                     // <= This should only be high when the `mult_pipe` is done.
      mult[done] = bin_read0_0.done;
      mult_pipe0.left = v_0.out;
      mult_pipe0.right = d_0.out;
      mult_pipe0.go = !mult_pipe0.done ? 1'd1;
    }
```

cgyurgyik:
Closes #369.
- Removes static delay of 1 for multi-cycle operations.
- Removed the constant static timing for `sqrt` and `div`. Given that the # iterations is based on the `WIDTH`, I don't think these hold.
- Does not slice or pad if the two values have equal width.


rachitnigam:
LGTM!

PR #371, opened 2021-05-04
Fix timing bug with multi-cycle.

cgyurgyik:
Writing to a register takes `1` cycle, and a `mult_pipe` takes 3 cycles for a total of 4 cycles. Thanks @rachitnigam for pointing out basic arithmetic to me.

Inadvertently fixes [Calyx 498](https://github.com/cucapra/calyx/issues/498).

### DIFF
--- a/src/main/scala/backends/futil/FutilBackend.scala
+++ b/src/main/scala/backends/futil/FutilBackend.scala
@@ -879,7 +879,7 @@ def emitInvokeDecl(app: EApp)(implicit store: Store, id2FuncDef: FunctionMapping
         // the multi-cycle operation is complete, if it exists.
         val (writeEnableSrcPort, delay) = out.multiCycleInfo match {
           case Some((compVar, Some(delay))) =>
-            (CompPort(compVar, "done"), out.delay.map(_ + delay))
+            (CompPort(compVar, "done"), out.delay.map(_ + delay + 1))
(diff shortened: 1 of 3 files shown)


rachitnigam:
Math! (LGTM)

src/main/scala/backends/futil/FutilBackend.scala
```diff
@@ -879,7 +879,7 @@ def emitInvokeDecl(app: EApp)(implicit store: Store, id2FuncDef: FunctionMapping
         // the multi-cycle operation is complete, if it exists.
         val (writeEnableSrcPort, delay) = out.multiCycleInfo match {
           case Some((compVar, Some(delay))) =>
-            (CompPort(compVar, "done"), out.delay.map(_ + delay))
+            (CompPort(compVar, "done"), out.delay.map(_ + delay + 1))
```
rachitnigam, line 882: Are we sure that this is always stored in a register? That’s the only time when adding 1 makes sense.

cgyurgyik: Yes, with this `CLet` expression, we're always writing to a register.

https://github.com/cucapra/dahlia/blob/1c88dc75284f13dd11887a20811d7203f0086ee4/src/main/scala/backends/futil/FutilBackend.scala#L870-L871

PR #386, opened 2021-10-14
Deprecate and remove Calyx fixed point with different widths.

cgyurgyik:
See: https://github.com/cucapra/calyx/issues/727

For now, different-width fixed point throws an error. 
If this is a blocker, we can implement it in the future.

Also fixes (expected) Calyx imports from an earlier PR.


rachitnigam:
Actually, I think you're right. For the computation for `x`, I was assuming that the backend or one of the lowering passes would add the right `as` expression to make the computation work out but that is not the case. You have to explicitly write it out.

In that case, LGTM!

PR #389, opened 2022-04-02
Start Scala 3 Migration

rachitnigam:
Scala 3 ships with a new tool called [Tasty][] that is able to read class file compiled in both Scala 2 and 3. This allows us to partially move some files to Scala 3 while keeping our hard Scala 2 dependency, the fastparse library, in Scala 2.

This PR uses the [migrate][] plugins and information from [mixed scala 2 and 3][scala-2-3] libraries.

[scala-2-3]: https://www.scala-lang.org/blog/2020/11/19/scala-3-forward-compat.html
[migrate]: https://docs.scala-lang.org/scala3/guides/migration/scala3-migrate.html
[tasty]: https://docs.scala-lang.org/scala3/guides/tasty-overview.html


sampsyo:
Very fancy!

PR #394, opened 2022-09-24
Map 1d arrays to Seq Mems

calebmkim:
Changed Calyx Backend to map 1d arrays to Sequential memories.

### DIFF
--- a/file-tests/should-futil/fixed-point-constant.expect
+++ b/file-tests/should-futil/fixed-point-constant.expect
@@ -1,4 +1,5 @@
 import "primitives/core.futil";
+import "primitives/memories.futil";
 import "primitives/binary_operators.futil";
 component main() -> () {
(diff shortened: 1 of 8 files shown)


rachitnigam:
Wowow! Quick as usual @calebmkim! The PR looks good but there is one challenge when changing Dahlia code: the Calyx tests will start failing. This is because the Calyx repo contains a bunch of tests that compile Dahlia programs to Calyx. When code generation changes in this repo are merged, Github Actions picks it up, tries to compile Dahlia programs to Calyx and complains when they no longer match.

Long story short, can you update the Dahlia tests in the Calyx repo by running `runt -i 'dahlia|examples`. Easiest way to do it is to merge this, then quickly open a new PR in the Calyx repo and see which tests fail. Then update and merge

PR #395, opened 2022-09-26
Don't rewrite top-level `decl` names

rachitnigam:
`decl` is used to bind external inputs. Previously, when `--lower`, the `decl` name would be rebound based on the number of dimensions of the array. This PR makes it so that the names are not rewritten and banking `decl` is disallowed. The overall effect is that `decl` names can be consistently used externally without being mangled.


rachitnigam:
This program is miscompiled:
```
decl alpha_int: ubit<32>[1];

for (let k: ubit<4> = 0..8) unroll 2 {
  let v: ubit<32> = alpha_int[0] * A[i][k];
}
```

rachitnigam:
The current problem is that we don't rewrite the indices when we access a `decl` memory on line 481 in lowerunroll

PR #396, opened 2022-10-11
always maps to seq-mems now

calebmkim:
I realized that the open PR I have in Calyx will only work if Dahlia maps to 2d, 3d, and 4d sequential mems. So I made some quick fixes to do that. 

Once we get this merged, I can update my current Calyx PR and see what fails.


rachitnigam (approved):
LGTM!

PR #397, opened 2022-10-20
Pass mutli-cycle info (needed to fix LeNet)

calebmkim:
We need to pass mutli-cycle info for the `EArrAccess` case for `EmitExpr()` in the Calyx backend. Before we weren't and that was creating incorrect Calyx code for LeNet. 

I also added a test case that would have caught the error I was running into in LeNet (basically for when Dahlia has something like`A[i] += value`.  When translated into Calyx we have a register to store the variable value, but that register's `write_en` signal needs to be `A.read_done`. Before it was just `1'd1`.)


rachitnigam:
LGTM!

PR #399, opened 2023-02-07
Rename Seq Mem Ports

calebmkim:
Small commit to prepare for https://github.com/cucapra/calyx/issues/1286

Also have the change ready for Calyx for when we push this.


sampsyo (approved):
Looks right to me!!

PR #400, opened 2023-05-04
Calyx Backend uses static groups instead of attribute

calebmkim:
Simple change to use `static group <lateny>` instead of `<"static"=latency>` for the Calyx backend.

### DIFF
--- a/file-tests/should-futil/fixed-point-constant.expect
+++ b/file-tests/should-futil/fixed-point-constant.expect
@@ -13,25 +13,21 @@ component main() -> () {
     fp_const3 = std_const(15,19968);
   }
   wires {
-    group let0<"static"=1> {
+    static group <1> let0 {
       a_0.in = fp_const0.out;
       a_0.write_en = 1'd1;
-      let0[done] = a_0.done;
     }
-    group let1<"static"=1> {
+    static group <1> let1 {
       b_0.in = fp_const1.out;
       b_0.write_en = 1'd1;
-      let1[done] = b_0.done;
     }
-    group let2<"static"=1> {
+    static group <1> let2 {
       c_0.in = fp_const2.out;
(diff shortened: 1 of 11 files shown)


file-tests/should-futil/fixed-point-constant.expect
```diff
@@ -13,25 +13,21 @@ component main() -> () {
...
   }
   wires {
-    group let0<"static"=1> {
+    static group <1> let0 {
       a_0.in = fp_const0.out;
       a_0.write_en = 1'd1;
-      let0[done] = a_0.done;
     }
-    group let1<"static"=1> {
+    static group <1> let1 {
```
rachitnigam, line 20: I'm confused, I thought the latency attribute comes after the name? Also, can we get rid of the space between the name and the latency?

rachitnigam (approved):
One clarification question but LGTM otherwise!

file-tests/should-futil/fixed-point-constant.expect
calebmkim: ah yeah my bad. I'll change it.

sampsyo:
Wahoo; nice work on this!

rachitnigam:
@calebmkim just checking what's the status of this PR?

calebmkim:
Ah, we can merge it. I was just thinking, though: without the `control-promotion` pass it will make very inefficient hardware: we will keep on instantiating a single fsm for each group. But maybe it'll be helpful just to check the correctness of our `compile-static` pass regardless?

rachitnigam:
Hm, that's a fair point! Should we wait till control promotion stuff is done and then merge it as a part of our efforts to close out the static groups milestone?

calebmkim:
Not going to merge this. We have decided to just emit `@promotable` as a hint that can be ignored.

PR #401, opened 2023-07-19
Fix Seq Mem Ports 

calebmkim:
Merging this will make the dahlia tests in https://github.com/cucapra/calyx/pull/1610 pass. 

We should probably wait until https://github.com/cucapra/calyx/pull/1610 is ready to merge, and then merge these at basically the exact same time.


sampsyo:
This all looks right from here!! I wish there were a good way to run the Calyx-side tests in CI when both sets of changes are in branches… such is the danger of separate repos, I guess.

Issue #403, opened 2024-01-08
Typechecker isSubtype does not handle alias type


Mark1626:
https://github.com/cucapra/dahlia/blob/master/src/main/scala/typechecker/Subtyping.scala#L44-L48

The following results in an error from the type checker

Case 1

```
record point { x: bit<32> }
def f(p: point): point = {
    let nextp: point = { x=p.x + 1 };
    return nextp;
}
```

```
Expected subtype of point in return, received: point
```

In this case `areEqual` receives the variable np as type `TRecType` while the return type is considered as `TAlias`.

---

Case 2

```
record point { x: ubit<32> }
let a: point = {x=10};
let b: point = (a as point);
```

The opposite scenario, `areEqual` receives `TAlias` for t1 and `TRecType` for t2.

PR #404, opened 2024-01-08
Handle TRecType in areEqual

Mark1626:
Fix for #403


rachitnigam:
LGTM! Thanks for contributing!

PR #406, opened 2024-01-12
Update documentation types and views

Mark1626:
Parts of the documentation seem to be incorrect/out-of-date. Updating the parts on types and views. Let me know if the updates are correct I'll edit if necessary

Summary of changes

## Cheat

1. Move the program structure section to the beginning. Felt this was a gocha when I first started trying out records and the compiler was failing.

2. Record literal, add semicolon at the end

3. Remove statement that says functions cannot have return types

4. Update with working examples in views section.

Note: I can't seem to lower views

---

## Types

1. Adding an example of `bit` and `ubit`

2. Floating Points example on auto inference is incorrect (bug?)

Compiling

```
let x = 1.0;
```

gives

```
// git.status = dirty, build.date = Sat Jan 06 22:05:58 IST 2024, git.hash = 8abb5e2
#include <ap_int.h>
extern "C" {
  void kernel() {
    
    #pragma HLS INTERFACE s_axilite port=return bundle=control
    double x = 1.0;
  }
}
```

3. Adding a message that calyx currently does not support float and double

4. Add section on fixed point

5. Add section on records

6. Add section on type casting

---

### DIFF
--- a/docs/cheat.md
+++ b/docs/cheat.md
@@ -6,6 +6,18 @@ title: Cheatsheet
 The next few sections will cover these in details. Feel free to skip this if
 nothing makes sense yet!
 
+## Program Structure
+
+A Dahlia program has the following structure:
+```
+<includes>
+<func defs or record defs>
+<decls>
+<commands>
+```
+
+Any one of these may be omitted but the order between them must be maintained.
+
 ## Constants
 
 ```dahlia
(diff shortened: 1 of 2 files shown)


Mark1626:
I think there should be another separate page on views

rachitnigam:
Thanks!

PR #409, opened 2024-01-17
BigInt in EInt to allow large literals in let

### LINKED ISSUE #408: Parse error in let expression for bitwidth 32 and above
Parsing the following will fail because `EInt` uses `Int` type

```
let x: ubit<32> = 0xffffffff;
```

Error message:

```
For input string: "ffffffff"
```

https://github.com/cucapra/dahlia/blob/f3735bcf6d249ee067fc9dc6fceb833f6bb1139e/src/main/scala/common/Syntax.scala#L138

**Suggestion:**

Changing `EInt` to use BigInt will fix the parse error. Subsequently `TStaticInt` will also have to use BigInt

https://github.com/cucapra/dahlia/blob/f3735bcf6d249ee067fc9dc6fceb833f6bb1139e/src/main/scala/common/Syntax.scala#L84

Mark1626:
Fixes #408 

Added new tests. Ran the existing unit test `sbt test` and integration tests with `runt`

### DIFF
--- a/src/main/scala/Parser.scala
+++ b/src/main/scala/Parser.scala
@@ -98,18 +98,18 @@ case class Parser(input: String) {
     positioned(
       P(
         "0" | "-".? ~ (CharIn("1-9") ~ CharsWhileIn("0-9").?)
-      ).!.map((n: String) => EInt(n.toInt)).opaque("integer")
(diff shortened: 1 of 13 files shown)


rachitnigam (approved):
LGTM! Thanks!

PR #412, opened 2024-01-29
Fix parsing issue of hex numbers in cast expr

### LINKED ISSUE #411: Parse issue for hexadecimal literals in cast expressions
The following fails with a strange message

```
let x = (0x9e3779b9 as ubit<32>);
```

Error message

```
[Error] For input string: "779b9 a" under radix 16
```

Issue seems to be the way hex literals are parsed

https://github.com/cucapra/dahlia/blob/ffb3866153e42f9073289e26ad243286cccbef98/src/main/scala/Parser.scala#L103-L113

A fix would be to handle it in the same way as Octal literals with the `CharsWhileIn("0-9a-fA-F")` instead of `CharIn("0-9a-fA-F").rep(1)`

Interestingly it seems to work on the online example editor in https://capra.cs.cornell.edu/dahlia/

Mark1626:
## Summary

1. Make the parser use the same logic as Octal literals
2. Add tests for this scenario

Closes #411

### DIFF
--- a/src/main/scala/Parser.scala
+++ b/src/main/scala/Parser.scala
@@ -102,7 +102,7 @@ case class Parser(input: String) {
     )
   def hex[K: P]: P[Expr] =
     positioned(
-      P("0x" ~/ CharIn("0-9a-fA-F").rep(1)).!.map((n: String) =>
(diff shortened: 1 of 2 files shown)


rachitnigam:
LGTM!

PR #419, opened 2024-03-03
Support unrolling functions with scalar arguments inside unrolled loops

### LINKED ISSUE #418: Question about using functions in unrolled loop
Dahlia currently does not allow functions to be present inside loops with unrolls.

While I understand that certain kinds of functions can cause an affine violation. I want to understand why Dahlia prevents it for all functions? 

### Simple functions defined on Records

```
record complex {
    real: float;
    imag: float
}

def complex_mul(a: complex, b: complex): complex = {
    let cr: float = (a.real * b.real) - (a.imag * b.imag);
    let ci: float = (a.real * b.imag) + (a.imag * b.real);
    let c: complex = { real=cr; imag=ci };
    return c;
}

let a: complex[1024 bank 4];
let b: complex[1024 bank 4];

let c: complex[1024 bank 4];

for (let i=0..1024) unroll 4 {
  c[i] := complex_mul(a[i], a[i]);
}
```

---

### Unrolling for imported functions

```
import vivado("math.h") {
  def cosf(v: float): float;
  def sinf(v: float): float;
}

let a: float[1024 bank 4];
let b: float[1024 bank 4];

for (let i=0..1024) unroll 4 {
  b[i] := sinf(a[i]);
}
```

In this case for the vivado backend during synthesis Vitis HLS generates multiple sincos units. Something similar for the Calyx backend?

Mark1626:
Closes #418 

Summary:
1. Implement the trait `Tracker` in `WFEnv`, to track function definitions
    * I've kept the key as `Id -> FuncDef`. This could have been `Id -> Boolean`, but it seemed too specific to this scenario for the `WFEnv`
2. Add a `canHaveFunctionInUnroll` function which will check from the `env` if the function can be present inside a unrolled loop
3. Update existing test case (which had scalar arguments) and add a new test case (with an array in the argument)

### DIFF
--- a/src/main/scala/common/Errors.scala
+++ b/src/main/scala/common/Errors.scala
@@ -88,7 +88,7 @@ object Errors {
       extends TypeError(s"$op cannot be inside an unrolled loop", pos)
 
   case class FuncInUnroll(pos: Position)
-      extends TypeError("Cannot call function inside unrolled loop.", pos)
+      extends TypeError("Cannot call function with non scalar arguments (like arrays) inside unrolled loop.", pos)
 
   // Unrolling and banking errors
   case class UnrollRangeError(pos: Position, rSize: Int, uFactor: Int)
--- a/src/main/scala/passes/WellFormedCheck.scala
+++ b/src/main/scala/passes/WellFormedCheck.scala
@@ -15,10 +15,37 @@ object WellFormedChecker {
   def check(p: Prog) = WFCheck.check(p)
 
   private case class WFEnv(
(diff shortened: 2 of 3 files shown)


rachitnigam:
Awesome, LGTM! Thanks @Mark1626

rachitnigam:
@Mark1626  BTW, it seems that the CI still doesn't run because of borked permissions. I am going give you write access to the repo so you can directly open branches within the repo and have the CI actually run.

PR #421, opened 2024-03-04
Migrate to Scala 3.3.1 LTS

Mark1626:
## Summary

1. Bumped up version of `scalatest`, `fastparse`, `sbtassembly`, `scalafmt`
2. Used the compiler opt `-new-syntax` along with `-rewrite` to format the code after a successful compile
3. Changes made in source `getClass.getResourceAsStream("/version.properties")` -> `scala.io.Source.fromResource("version.properties")`
4. **The compiler noted that in certain cases where for an `Option[T]` type only `Some[T]` was consumed and `None` was ignored**. This has been temporarily marked with `@unchecked`
5. Fatjar construction had a considerable change as `sbtassembly` went up from `0.14.9` to `2.1.5` in `build.sbt`
6. The Dotty Compiler cleared minor issues in pattern matching that were marked eg)`Parser.scala L418`, `calyx/Backend.scala`
7. Main function still follows Scala 2 pattern rather than a Scala 3 `@main` approach
8. Tested with the existing unit tests and functional tests
9. The documentation is out of date, should I do this in this PR itself.

Note: Most of this was just a straightforward version update except for point 4 and 5

--- 

Let me know your feedback, I'll work on them.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,2 +1,2 @@
-version = "2.4.2"
+version = "3.8.0"
 align.tokens = []
--- a/build.sbt
+++ b/build.sbt
@@ -1,12 +1,13 @@
 name := "Dahlia"
 version := "0.0.2"
 
-scalaVersion := "2.13.12"
+scalaVersion := "3.3.1"
 
 libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.0.8" % "test",
+  "org.scalatest" %% "scalatest" % "3.2.18" % "test",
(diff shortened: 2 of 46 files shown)


Mark1626:
Any comments on point 4

rachitnigam:
How about (once this is merged), opening a new PR immediately where the `@unchecked` is remove so we can stare at the error message and see which places are obviously incorrect and which places can be handled directly

Mark1626:
Sounds good. Do you want me to merge it to master?

rachitnigam:
Yup! You should have access to hit the green button now!

rachitnigam:
I generally prefer squash commits BTW

PR #429, opened 2024-04-01
Add `Repeat` to Calyx Backend

calebmkim:
Instead of compiling to `@bound(n) while` the Calyx backend will compile to `repeat n` statements.


rachitnigam:
Awesome! LGTM!

rachitnigam:
FYI: The new changes will not be picked up by the Calyx CI by default. A separate PR in that repo is needed to bump the version number

Issue #427, opened 2024-03-21
Calyx back-end issues


Mark1626:
I noted these two issue when I was trying to add float support in Dahlia (Calyx backend)

## 1.  Hoist memory pass issue

The following does not pass with `--lower` flag.

```
decl a: float[128];
decl b: bit<10>[128];
decl c: float[128];

for (let i:bit<8>=0..128) {
    c[i] := a[b[i]];
}
```

**Error**

```
[Type error] [Line 0, Column 0] `b_read0' is not bound in scope.
<undefined position>

```

Root cause seems to be the Hoist pass, `b_read0` has not been created

```
===============Original===============
decl a: float[128];
decl b: bit<10>[128];
decl c: float[128];
for (let i: bit<8> = 0 .. 128) {
  c[i] := a[b[i]];
}
======================================
===============Hoist memory reads===============
decl a: float[128];
decl b: bit<10>[128];
decl c: float[128];
for (let i: bit<8> = 0 .. 128) {
  let a_read0 = a[b_read0];
  c[i] := a_read0;
}

```

---

## 2. Incorrect usage of signed Calyx primitives

The following generates Calyx IR

```
decl a: bit<10>[128];
decl b: bit<10>[128];
decl c: bit<10>[128];

for (let i:bit<8>=0..128) {
    c[i] := a[i] & b[i];
}

```

The Calyx IR contains a `std_sand` primitive which seems to be not present in the current Calyx primitives 

```
import "primitives/core.futil";
import "primitives/memories/seq.futil";
import "primitives/binary_operators.futil";
import "P32.futil";
component main() -> () {
  cells {
    @external(1) a = seq_mem_d1(10,128,8);
    a_read0_0 = std_reg(10);
    add0 = std_sadd(8);
    and0 = std_sand(10);
   ...
   ...
   ...
```

rachitnigam:
Thanks! (1) seems like a bug in lowering while (2) is probably in the backend. For (2), we basically attempt to generate the string name of the operator based on whether or not the operands are signed or not (adding the `s` if they are). For logical operators like `&`, I think we should probably not generate the `s`?

Mark1626:
2. Seems to be because of this function. I feel the Calyx backend code needs some restructuring, we seem to lose information about the operator

https://github.com/cucapra/dahlia/blob/9ec9a58fa5fefc4e3fc5e19a38d90c3f780ec4b8/src/main/scala/backends/calyx/Ast.scala#L366-L370

[...]

PR #430, opened 2024-04-09
Fix Calyx backend issue with emitting signed binops

Mark1626:
This fixes point number 2 in #427 

## Background

Dahlia was emitting Calyx primitives incorrectly for signed ops.

eg)

```
decl a: bit<10>[128];
decl b: bit<10>[128];
decl c: bit<10>[128];

for (let i:bit<8>=0..128) {
    c[i] := a[i] & b[i];
}

```

The Calyx IR contains a std_sand primitive

## Changes Summary

1. Modify the function `signed` in Helpers.scala to use the op to decide if a given op is signed or not
   1. Match for op type `NumOp` and use the previous behaviour
   2. Any other op type binary operator primitives are not signed eg) bitwise, comparison
3. Add an test with signed bit and binops for regression

### DIFF
--- a/file-tests/should-futil/emit-signed-op.expect
+++ b/file-tests/should-futil/emit-signed-op.expect
@@ -1,0 +1,110 @@
+import "primitives/core.futil";
+import "primitives/memories/seq.futil";
+import "primitives/binary_operators.futil";
+component main() -> () {
+  cells {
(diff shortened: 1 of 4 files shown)


rachitnigam:
LGTM!

